Issue #2, opened 2026-07-28
Milestone v1.0.1 - First Buildable Core


Ayush-Patel-56:
# Milestone v1.0.1 - First Buildable Core

| | |
|---|---|
| **Project** | Trajectory IR |
| **Milestone** | v1.0.1 - Phase 1A completion |
| **Purpose of this file** | Step-by-step execution guide for the first buildable milestone: environment, dependencies, build order, code, verification, and timeline. This is an execution runbook, not a spec - for the normative design, see `TRAJECTORY_IR_MASTER_SPEC.md` / `README.md`. |
| **Audience** | Whoever is writing code this milestone - human or AI coding agent. |
| **Status** | Ready to execute. |

---

## 0. Aim

A working, DBOS-backed Trajectory IR core where R01 and R02 pass against real crash-and-resume behavior, demonstrated end-to-end by a kill-mid-deploy script. This is the milestone where the spec stops being a document and becomes a thing that provably does what §1 of the master spec claims.

**Definition of done (all four required, no partial credit):**

1. `pip install -e .` works from a clean clone.
2. `pytest conformance/r01_seal_resume_test.py conformance/r02_non_idempotent_test.py` passes.
3. `python examples/kill-mid-deploy/run_demo.py`, followed by a real `kill -9` and a restart, produces exactly one execution of the deploy tool and zero extra model calls.
4. Everything committed with DCO sign-off, CI green.

**Estimated time:** 3–4 weeks at a realistic student-team pace (a few focused hours most days, not full-time). Per-task breakdown in §9. If you move faster, good - but don't skip the verification steps to hit a date. A milestone that "passes" without the crash-injection test actually being adversarial is worse than being a week late.

---

## 1. A gap carried in from spec review - read this before writing code

The master spec's §8.0/§8.1 say `DECISION` and `TOOL_CALL` are wrapped as the durable-execution backend's own step primitive. They do not say the **model inference call itself** must also be wrapped as a step. This matters concretely: DBOS workflow functions replay their entire body from the top on recovery - only calls wrapped as `@DBOS.step()` get their result memoized and skipped. An unwrapped model call sitting in the workflow body will be **re-invoked in full on every resume**, even after `DECISION` is already sealed; the fresh (and wasted) output just gets silently discarded once execution reaches the memoized decision-write step.

[...]

PR #8, opened 2026-07-30
Milestone v1.0.1: First Buildable Core

Ayush-Patel-56:
Implements the DBOS-backed, crash-safe Trajectory IR execution core described in the milestone v1.0.1 design spec and plan (tasks in #2).

## What this adds

- Content-addressed `Node` model with RFC 8785 hashing, so a node's identity is order-independent and timestamp-invariant.
- A SQLite-backed, idempotent, append-only `NodeLog` (`INSERT OR IGNORE` on content id) that seals every step of a trajectory.
- Fail-closed `EffectClass` classification from MCP tool annotations, so an ambiguous or contradictory annotation is treated as non-idempotent rather than assumed safe.
- A DBOS durable-backend adapter wrapping model inference and tool calls as steps/workflows, so a crash mid-run can resume from durable state instead of replaying from scratch.
- A block-and-gate mechanism for non-idempotent tool calls (`BlockedNeedsGate`): if a tool call was already attempted for a given step and sequence, a resume blocks re-entry instead of silently re-running a side effect like a deploy.
- A durable `run_step` workflow that wires inference, decision sealing, and gated tool execution together, plus a small client SDK (`open_trajectory`, `project`, `seal_decision`, `exec_tool`, `commit_step`, `resume`) over the same primitives.
- A `kill-mid-deploy` fixture agent and demo (`examples/kill-mid-deploy/`) that can be killed mid-flight at three different points and resumed, to make the crash-safety claim concrete rather than theoretical.
- Two conformance tests under `conformance/` tied to the milestone's Definition of Done: R01 (model inference is not reinvoked after a decision is sealed) and R02 (a crash mid non-idempotent tool call blocks on resume instead of retrying).
- A CI workflow running the unit, e2e, and conformance suites on push to main and on pull requests.

## Notable implementation decisions

A couple of places where the plan's reference code had a real gap were fixed rather than transcribed as-is, each time confirmed with the repo owner before changing course:

- The tool-call gate originally blocked only when a `TOOL_CALL` node was present and no `TOOL_RESULT` was present yet. That fails open if a crash happens after the result is logged but before the workflow step is durably memoized, and it can also let one tool's result mask a different interrupted tool's missing result in a multi-tool step. The gate now blocks on the presence of `TOOL_CALL` at that exact step/sequence alone, which is sufficient proof of a prior attempt.
- The client SDK's `exec_tool` and `commit_step` originally used hardcoded sequence numbers, so two non-idempotent tool calls in the same step would collide and the second call would be falsely blocked as a duplicate on its very first attempt. Both now take a caller-supplied `seq`.

## Test plan

- [x] Unit tests pass (`pytest test/unit -v`)
- [x] E2E crash/resume scenarios pass (`pytest test/e2e -v`)
- [x] Conformance suite passes (`pytest conformance/ -v`)
- [x] Manual kill -9 + resume of the kill-mid-deploy demo verified end to end
- [ ] CI green on this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - run: pip install -e ".[dev]"
+      - run: pytest test/unit -v
+      - run: pytest test/e2e -v
+      - run: pytest conformance/ -v
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,9 @@ venv.bak/
 # OS generated files
 .DS_Store
 Thumbs.db
+
+# Build and runtime artifacts
+*.sqlite
+test_model_call_count.txt
+test_deploy_side_effect_count.txt
+*.marker
--- a/client/python/trajectory_client.py
+++ b/client/python/trajectory_client.py
@@ -52,11 +52,24 @@ def seal_decision(trajectory: Trajectory, step_n: int, plan: dict) -> Decision:
     return Decision(step_n=step_n, plan=plan)
 
 
-def exec_tool(trajectory: Trajectory, step_n: int, call: dict, tool: Tool) -> ToolResult:
+def exec_tool(trajectory: Trajectory, step_n: int, call: dict, tool: Tool, seq: int) -> ToolResult:
+    """Execute a tool call within a step.
+
+    Args:
+        trajectory: The trajectory context
+        step_n: Step number
+        call: Tool call dict with "args" key
+        tool: Tool definition with name, fn, and effect_class
+        seq: Sequence number within the step (caller-supplied, must be unique per call)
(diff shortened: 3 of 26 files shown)


Siddhartha-singh01:
Thanks for this  solid Phase 1A core. R01/R02 + real kill/resume and the gate
fixes (TOOL_CALL-at-seq, caller-supplied seq) are exactly the bar we care about.
CI is green on the test job; DCO trailers look good on the commits I checked.

Before / right after merge, a few owner notes:

1. Layout: this PR uses drivers/durable_backend (underscore). Master README and
   our scaffold PR used durable-backend. We should pick one name so we don't
   fork the tree.
2. CI here runs unit/e2e/conformance only. We still want DCO + lint as hard
   checks (can be a fast follow-up PR; don't need to block the whole milestone
   if tests stay green).
3. Phase-0 files under spec/ aren't in this PR. Happy to land them in a small
   docs follow-up so "spec before code" stays true for new contributors.
4. docs/superpowers/ is fine as working notes  please treat root README as
   normative if anything conflicts.

I'm treating this PR as the mainline for v1.0.1 core. Scaffold/process PRs that
overlap will be closed or rebased as follow-ups rather than merged on top as-is.

[Pick one:]
- Request: address (1) path naming in this PR if it's a quick rename, then LGTM.
- Or: LGTM on core; I'll merge and open a process/layout follow-up.

PR #53, opened 2026-07-31
feat: R04 default context projector with CONSTRAINT budget safety

### LINKED ISSUE #52: feat: R04 context projector with CONSTRAINT budget safety
## Problem

Runnable conformance so far:

| ID | Status |
|----|--------|
| R01 / R02 | Done (crash/resume + block-and-gate) |
| R03 | Done (PURE recompute, #44) |
| R05 | Done (`.tir` hash round-trip, #42) |

**R04** is the next §10 item that follows this path:

> R04  A `CONSTRAINT` node is never silently dropped under budget pressure; the system either includes it or raises a hard error.

Today `client.project()` / `PROJECT_CONTEXT` only **records** a caller-supplied context dict. There is no context **projector** that:

1. Selects nodes for the next model call under a token (or size) budget
2. Guarantees every `CONSTRAINT` is kept (or fails hard)
3. Implements the default policy from README §9 / §7 note

Without this, R04 cannot be a real conformance gate, and PROJECT_CONTEXT is not yet the semantic assembly step the step lifecycle describes (§8.1).

## Proposal

Implement a **minimal default context projector** and a named R04 conformance suite.

### Goals

1. **Python** (reference first, same order as nodes / resume / `.tir`):
   - Module under `pkg/trajectory_ir/` (e.g. `runtime/projector.py` or `projector/`) with something like:
     - `project_context(nodes, *, budget_tokens: int, pinned_ids: set[str] | None = None) -> dict`
     - or `ProjectResult` with included node ids + assembled payload
   - **Default policy** (when no `projector-policy.yaml`):
     - Always include every `CONSTRAINT` node
     - Always include explicitly pinned node ids (if provided)
     - If constraints alone exceed the budget → raise hard error (e.g. `BudgetImpossible` / `BUDGET_IMPOSSIBLE`), never silent drop
     - Other node kinds may be trimmed by a simple, deterministic rule (document it; keep naive/v0.1-sized)
   - Wire optionally into `client.project` or keep as a pure function called by the client (prefer pure function + thin client call so tests stay easy)
2. **Conformance:** `conformance/r04_constraint_budget_test.py`
   - Constraints fit under budget → all CONSTRAINT ids present in projected set
   - Constraints alone exceed budget → hard error, no silent partial drop of constraints
   - Optionally: non-constraint nodes can be dropped when budget is tight, without dropping constraints
3. **README §10:** mark R04 runnable (same style as R03/R05)
4. **Go** (if low cost after Python green): port the same pure projector + unit tests under `go/trajir/…`; not required to block Python merge if scope grows, but prefer parity when the surface is small
5. CHANGELOG note

### How it connects

1. Uses existing node kinds (`CONSTRAINT`, etc.) and NodeLog / list of node records
2. Completes the step-lifecycle story after PROJECT_CONTEXT is more than a pass-through append
3. Does **not** require package signatures, Fluid, Restate, or multi-writer HA

## Spec reference

1. Root README §6.2 (`CONSTRAINT`), §8.1 step lifecycle, §9 default projector policy note, §10 R04
2. Current pass-through: `client/python/trajectory_client.py` `project()`, Go `client.Project`

## Explicitly not part of this idea

1. Full custom `projector-policy.yaml` parser / operator DSL (default built-in policy only for this issue)
2. Real LLM tokenizers / model-specific counting  a simple, documented size metric is enough for R04 (e.g. JSON char budget or node-count budget); document the metric
3. R06 sandbox, R07 graft, R08 context redaction (export redaction already exists)
4. Package signatures, k8s/Fluid, second durable backend

## Acceptance

1. Projector never silently drops a `CONSTRAINT` under budget pressure
2. `BUDGET_IMPOSSIBLE` (or equivalent) when constraints alone exceed budget
3. `pytest conformance/r04_constraint_budget_test.py` (name may vary) passes in CI Quality
4. R01–R03 and R05 still green
5. README §10 lists R04 as runnable
6. DCO; PR follows project template

## Parent / sequence

Follows the conformance track:

R01 → R02 → R05 → R03 → **R04 (this issue)** → later R06–R08.

Phase 1A bar already met; this is tracked §10 work after dual-language packages and resume matrix hardening.

Siddhartha-singh01:
## Summary
Implements **R04**: CONSTRAINT nodes are never silently dropped under budget pressure. Closes #52.

### Python
- `pkg/trajectory_ir/runtime/projector.py` — `project_context`, `BudgetImpossible`, `ProjectResult`
- Metric: `json_chars` (documented; not a real LLM tokenizer)
- Default policy: always include CONSTRAINT + pinned; hard fail if must-include exceeds budget; optional fill of other kinds
- Unit + `conformance/r04_constraint_budget_test.py`
- Client `project()` docstring points at the pure projector

### Go
- `go/trajir/projector` with matching API + tests
- Documented in `go/README.md`

### Docs
- README §10 R04 runnable; CHANGELOG

## Test plan
- [x] `pytest test/unit/test_projector.py conformance/`
- [x] `go build/vet ./trajir/projector` (local Windows may block package test binaries; CI Linux authoritative)
- [ ] CI green

### DIFF
--- a/go/trajir/projector/projector.go
+++ b/go/trajir/projector/projector.go
@@ -1,15 +1,20 @@
 // Package projector implements the default context projector (R04).
 //
-// Metric json_chars: UTF-8 length of compact sort-keyed JSON
-// {"kind":...,"payload":...} per node. Matches Python
-// pkg/trajectory_ir/runtime/projector.py.
+// Metric rfc8785_bytes: length of RFC 8785 (JCS) canonical form of
+// {"kind":...,"payload":...} per node. Same JCS stack as trajir/nodes
+// (github.com/gowebpki/jcs) so size units match Python rfc8785.
 package projector
 
 import (
 	"encoding/json"
 	"fmt"
 	"sort"
+
+	"github.com/gowebpki/jcs"
 )
+
+// SizeMetric is the documented budget unit name (parity with Python SIZE_METRIC).
+const SizeMetric = "rfc8785_bytes"
 
 // BudgetImpossible is raised when CONSTRAINT/pinned nodes alone exceed budget.
 type BudgetImpossible struct {
@@ -35,19 +40,23 @@ type Result struct {
 	Metric      string
 }
 
-// NodeSizeUnits returns the json_chars size of one node record.
+// NodeSizeUnits returns the rfc8785_bytes size of one node record (JCS length).
 func NodeSizeUnits(node map[string]any) (int, error) {
 	kind, _ := node["kind"].(string)
 	payload := node["payload"]
 	if payload == nil {
 		payload = map[string]any{}
 	}
+	// encoding/json then jcs.Transform — same pattern as nodes.PayloadHash.
 	raw, err := json.Marshal(map[string]any{"kind": kind, "payload": payload})
 	if err != nil {
-		return 0, err
+		return 0, fmt.Errorf("json marshal: %w", err)
 	}
-	// encoding/json sorts map keys; matches Python sort_keys=True for this shape.
-	return len(raw), nil
+	canon, err := jcs.Transform(raw)
+	if err != nil {
+		return 0, fmt.Errorf("jcs: %w", err)
+	}
+	return len(canon), nil
 }
 
 type sortKey struct {
@@ -192,11 +201,11 @@ func ProjectContext(nodes []map[string]any, budget int, pinnedIDs map[string]str
 		})
 	}
 	ctx := map[string]any{
-		"items":         items,
-		"included_ids":  ids,
-		"budget":        budget,
-		"size_units":    used,
-		"metric":        "json_chars",
+		"items":        items,
+		"included_ids": ids,
+		"budget":       budget,
+		"size_units":   used,
+		"metric":       SizeMetric,
 	}
 	return &Result{
 		IncludedIDs: ids,
@@ -204,6 +213,6 @@ func ProjectContext(nodes []map[string]any, budget int, pinnedIDs map[string]str
 		Context:     ctx,
 		SizeUnits:   used,
 		Budget:      budget,
-		Metric:      "json_chars",
+		Metric:      SizeMetric,
 	}, nil
 }
--- a/go/trajir/projector/projector.go
+++ b/go/trajir/projector/projector.go
@@ -1,0 +1,209 @@
+// Package projector implements the default context projector (R04).
+//
+// Metric json_chars: UTF-8 length of compact sort-keyed JSON
+// {"kind":...,"payload":...} per node. Matches Python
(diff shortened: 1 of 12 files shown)


pkg/trajectory_ir/runtime/projector.py
```diff
@@ -0,0 +1,157 @@
...
+    budget: int
+    metric: str = "json_chars"
+
+
+def node_size_units(node: dict[str, Any]) -> int:
+    """Size of one node under the json_chars metric."""
+    kind = node.get("kind", "")
+    payload = node.get("payload", {})
+    raw = json.dumps({"kind": kind, "payload": payload}, sort_keys=True, default=str)
+    return len(raw.encode("utf-8"))
```
Ayush-Patel-56: Change `node_size_units` (Python) and `NodeSizeUnits` (Go) to use the same RFC 8785 canonicalization already used in `nodes.py`/`nodes.go`, instead of raw `json.dumps`/`json.Marshal`

go/trajir/projector/projector.go
```diff
@@ -0,0 +1,209 @@
...
+	if payload == nil {
+		payload = map[string]any{}
+	}
+	raw, err := json.Marshal(map[string]any{"kind": kind, "payload": payload})
+	if err != nil {
+		return 0, err
+	}
+	// encoding/json sorts map keys; matches Python sort_keys=True for this shape.
+	return len(raw), nil
+}
```
Ayush-Patel-56, line 60: add `import rfc8785` at the top, drop the old `json.dumps(..., sort_keys=True, default=str).encode("utf-8")` line

Ayush-Patel-56 (changes requested):
add `"github.com/gowebpki/jcs"` import, same pattern as `go/trajir/nodes/nodes.go:74-78`

That makes size units byte-identical between Python and Go, closing the `BUDGET_IMPOSSIBLE` divergence. I'd also ask to add one test that feeds the same node set through both and asserts equal `size_units`.

Siddhartha-singh01:
Addressed review (CHANGES_REQUESTED):

1. **Python** \`node_size_units\` now uses \`rfc8785.dumps\` (same as node hashing).
2. **Go** \`NodeSizeUnits\` now uses \`json.Marshal\` + \`github.com/gowebpki/jcs\` (same pattern as \`trajir/nodes\`).
3. Metric renamed to \`rfc8785_bytes\` for honesty.
4. **Cross-language goldens:** \`testdata/projector_size_vectors.json\` + Python \`test_projector_size_vectors.py\` + Go \`TestProjectorSizeVectorsMatchPythonGoldens\`.

Ready for re-review / merge.

Ayush-Patel-56 (approved):
lgtm

PR #99, opened 2026-08-06
fix: Go effect classifier fails closed on openWorldHint

### LINKED ISSUE #98: Go effect classifier is missing the openWorldHint fail-closed check
## Summary

`go/trajir/effects/effects.go` does not match the Python reference implementation in `pkg/trajectory_ir/effects/classify.py`. The Go version is missing one of the four conditions required before a tool can be classified `IDEMPOTENT_WRITE`, which means it can end up less strict than Python for the same input.

## Where

- Python: `pkg/trajectory_ir/effects/classify.py` lines 42-49
- Go: `go/trajir/effects/effects.go` line 41

## Details

Python requires `openWorldHint` to not be `True` before classifying a tool as `IDEMPOTENT_WRITE`:

```python
if (
    annotations.get("readOnlyHint") is False
    and annotations.get("idempotentHint") is True
    and annotations.get("destructiveHint") is False
    and annotations.get("openWorldHint") is not True
):
    # openWorld + write is treated as non-idempotent unless operator overrides.
    return EffectClass.IDEMPOTENT_WRITE
```

Go only checks the first three conditions and never reads `openWorldHint`:

```go
if hasReadOnly && !readOnly && hasIdempotent && idempotent && hasDestructive && !destructive {
    return IDEMPOTENT_WRITE
}
```

The Go file's own doc comment says "Rules match pkg/trajectory_ir/effects/classify.py (fail closed)," so this is a real divergence from what the code claims to do, not an intentional difference.

## Why it matters

Given the annotation payload:

```json
{
  "readOnlyHint": false,
  "idempotentHint": true,
  "destructiveHint": false,
  "openWorldHint": true
}
```

Python correctly falls through to `NON_IDEMPOTENT_WRITE` and gates the tool. Go classifies the same payload as `IDEMPOTENT_WRITE`, so `RequiresBlockAndGate` returns `false` and the tool skips the atomic `ClaimToolCall` / NodeLog gate on crash-resume.

That gap matters most for open world tools with real, un-deduplicated side effects (sending a payment, placing an order, posting to a third party API). A tool server that self-reports `idempotentHint: true` for one of these gets treated as safe by the Go runtime and unsafe by the Python one, for identical input.

`pkg/effects/` is one of the two paths SECURITY.md calls out as requiring maximum scrutiny before merge, so this is worth fixing directly rather than deferring.

## Secondary note

`durable.Step` in `go/trajir/durable/backend.go` runs the step function and saves the memo without holding a lock across the whole sequence, so two concurrent calls for the same workflow/step key can both run the function before either saves. On its own this isn't the actual safety boundary, since `doc.go` states the NodeLog is the source of truth for gating `NON_IDEMPOTENT_WRITE`, not backend internals. But if a tool is misclassified due to the bug above, it takes this ungated path, and the two issues compound into a real double-execution risk. Worth a look while this is being fixed, even if it doesn't need its own issue.

## Suggested fix

- Add the `openWorldHint` conjunct to the Go `IDEMPOTENT_WRITE` branch so it matches `classify.py` exactly.
- Add a conformance test that runs the same set of annotation fixtures through both `classify_from_mcp` and `ClassifyFromMCP` and asserts identical output, so this kind of drift gets caught automatically instead of by manual review.

## Affected files

- `go/trajir/effects/effects.go`
- `go/trajir/effects/effects_test.go` (no existing coverage for `openWorldHint`)

jakharmonika364:
## Summary

Fixes #98.

`ClassifyFromMCP` in `go/trajir/effects/effects.go` was missing the `openWorldHint` check that `classify.py` has. An open world tool that self-reports `idempotentHint: true` could get classified `IDEMPOTENT_WRITE` in Go, skipping the block-and-gate NodeLog claim on resume, while the same annotations correctly stay `NON_IDEMPOTENT_WRITE` in Python.

## Change

- Added the `openWorldHint` conjunct to the `IDEMPOTENT_WRITE` branch so it matches the Python rule: `openWorldHint` must not be `true` for a tool to qualify as `IDEMPOTENT_WRITE`.
- Added two tests: one confirming `openWorldHint: true` fails closed to `NON_IDEMPOTENT_WRITE` even with idempotent/non-destructive/non-readonly hints set, and one confirming `openWorldHint: false` still allows `IDEMPOTENT_WRITE` as before.

## Test plan

- [x] `go test ./trajir/effects/...` passes, including the two new cases
- [x] Existing classifier tests still pass, no behavior change for inputs that don't set `openWorldHint`

### DIFF
--- a/go/trajir/effects/effects.go
+++ b/go/trajir/effects/effects.go
@@ -34,11 +34,13 @@ func ClassifyFromMCP(annotations map[string]any) EffectClass {
 	readOnly, hasReadOnly := asBool(annotations["readOnlyHint"])
 	destructive, hasDestructive := asBool(annotations["destructiveHint"])
 	idempotent, hasIdempotent := asBool(annotations["idempotentHint"])
+	openWorld, hasOpenWorld := asBool(annotations["openWorldHint"])
 
 	if hasReadOnly && readOnly && !(hasDestructive && destructive) {
 		return READ_ONLY
 	}
-	if hasReadOnly && !readOnly && hasIdempotent && idempotent && hasDestructive && !destructive {
+	// openWorldHint=true without a clear safe profile stays fail-closed below.
+	if hasReadOnly && !readOnly && hasIdempotent && idempotent && hasDestructive && !destructive && !(hasOpenWorld && openWorld) {
 		return IDEMPOTENT_WRITE
 	}
 	return NON_IDEMPOTENT_WRITE
--- a/go/trajir/effects/effects_test.go
+++ b/go/trajir/effects/effects_test.go
@@ -57,6 +57,30 @@ func TestExplicitIdempotentWrite(t *testing.T) {
 	}
 }
 
+func TestOpenWorldFailsClosedEvenIfIdempotent(t *testing.T) {
+	got := effects.ClassifyFromMCP(map[string]any{
+		"readOnlyHint":    false,
+		"idempotentHint":  true,
+		"destructiveHint": false,
+		"openWorldHint":   true,
+	})
+	if got != effects.NON_IDEMPOTENT_WRITE {
+		t.Fatalf("got %s, want NON_IDEMPOTENT_WRITE", got)
+	}
+}
+
+func TestOpenWorldFalseStillAllowsIdempotentWrite(t *testing.T) {
+	got := effects.ClassifyFromMCP(map[string]any{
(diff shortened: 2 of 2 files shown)


Siddhartha-singh01 (approved):
Looks good. Small Go parity fix so openWorldHint matches Python and fails closed. Tests cover the true/false cases cleanly. Approving and merging.

PR #100, opened 2026-08-06
fix: log TOOL_CALL and TOOL_RESULT nodes for non-gated tool calls

### LINKED ISSUE #89: fix: client SDK exec_tool does not log TOOL_CALL/TOOL_RESULT for non-gated tools
## Problem

`exec_tool` in `client/python/trajectory_client.py` (line 85-118) only appends `TOOL_CALL`/`TOOL_RESULT` nodes to the log for `NON_IDEMPOTENT_WRITE` tools, through `make_gated_tool_call`. For every other effect class (`PURE`, `READ_ONLY`, `IDEMPOTENT_WRITE`, `AGENT_SPAWN`) it takes the `else` branch at line 116, sets `fn = tool.fn`, and calls it directly at line 117. Nothing gets written to the NodeLog for that call.

Ran `examples/host_loop/run_host.py` to confirm this in practice: a `greet` call (PURE) leaves no trace in the trajectory log at all, only the `deploy_server` call (NON_IDEMPOTENT_WRITE) shows up as a TOOL_CALL/TOOL_RESULT pair.

This is basically the same bug #45 fixed, just in a different place. #45 fixed `resume/step.py`'s `run_step`, which now logs plain tools correctly (see the comment around line 141, "Plain tools still need IR history audit, gate path already logs"). The client SDK's `exec_tool` never got that fix, it has its own separate tool-dispatch path that duplicates what `run_step` does and missed the same case.

None of the existing tests catch this because every test that checks NodeLog contents through `exec_tool` happens to use a NON_IDEMPOTENT_WRITE tool: `test/unit/test_client_sdk.py`, `conformance/r06_sandbox_test.py`, `test/unit/test_host_loop_example.py`.

## Fix

Make the non-gated branch in `exec_tool` append TOOL_CALL/TOOL_RESULT the same way `run_step`'s plain-tool branch does, or have `exec_tool` call through the same shared logic instead of duplicating tool dispatch. Add a NodeLog assertion for a PURE or READ_ONLY tool call in `test_client_sdk.py` and `test_host_loop_example.py` so this can't regress silently again.

## Spec reference

README §4 principle 1, §8.1 step 4. Related: #45

jakharmonika364:
## Problem

`exec_tool` in `client/python/trajectory_client.py` only appended TOOL_CALL and TOOL_RESULT nodes on the block-and-gate path, which only covers NON_IDEMPOTENT_WRITE calls. Anything classified PURE, READ_ONLY, or IDEMPOTENT_WRITE ran through `fn(**call["args"])` directly with no node log write at all, so those calls left no trace in the IR history.

## Fix

The non-gated branch now appends TOOL_CALL at `seq` and TOOL_RESULT at `seq + 1`, mirroring the slot layout the gated path already uses in `resume/step.py`. Every tool call is now represented in the log regardless of effect class.

## Testing

Added `test/unit/test_client_sdk_plain_tool_logging.py` covering a single non-gated call and two non-gated calls in the same step with distinct seq slots. Full `test/unit` suite passes locally (98 passed, 2 skipped, unrelated to this change).

Closes #89

### DIFF
--- a/client/python/trajectory_client.py
+++ b/client/python/trajectory_client.py
@@ -112,9 +112,27 @@ def exec_tool(trajectory: Trajectory, step_n: int, call: dict, tool: Tool, seq: 
             tool_name=tool.name,
             tool_fn=tool.fn,
         )
+        result = fn(**call["args"])
     else:
-        fn = tool.fn
-    result = fn(**call["args"])
+        # Not claim-gated, but still needs IR history audit; gate path
+        # already logs its own TOOL_CALL/TOOL_RESULT nodes.
+        result = tool.fn(**call["args"])
+        log.append(
+            "TOOL_CALL",
+            step_n,
+            {"tool": tool.name, "args": call["args"]},
+            trajectory.trajectory_id,
+            trajectory.tenant_id,
+            seq=seq,
+        )
+        log.append(
+            "TOOL_RESULT",
+            step_n,
+            {"result": result},
+            trajectory.trajectory_id,
+            trajectory.tenant_id,
+            seq=seq + 1,
+        )
     return ToolResult(step_n=step_n, result=result)
 
 
--- a/test/unit/test_client_sdk_plain_tool_logging.py
+++ b/test/unit/test_client_sdk_plain_tool_logging.py
@@ -1,0 +1,36 @@
+from client.python.trajectory_client import exec_tool, open_trajectory
+from trajectory_ir.effects import EffectClass
+from trajectory_ir.runtime.log import NodeLog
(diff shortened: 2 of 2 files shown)


Siddhartha-singh01 (approved):
Good fix for the plain tool path leaving no IR history. Logging TOOL_CALL/TOOL_RESULT for non gated tools matches what people would expect from the log. Soft note for later: logging after the tool returns means a crash mid tool still leaves no trail, but that can be a follow up. Approving; will merge carefully relative to #107 since both touch trajectory_client.py.

PR #101, opened 2026-08-06
fix: force CONSTRAINT into always_include_kinds on every policy

### LINKED ISSUE #90: fix: CONSTRAINT nodes can be silently dropped by a custom projector policy
## Problem

`ProjectorPolicy.__post_init__` in `pkg/trajectory_ir/runtime/policy.py` (line 45-53) only checks that `always_include_kinds` is non-empty, nothing requires `CONSTRAINT` to actually be a member of it. `policy_from_mapping` (line 61-85) accepts whatever kind list a policy file supplies without checking that either.

In `project_context` (`pkg/trajectory_ir/runtime/projector.py`, around line 109-144), any node kind that isn't in `always_include_kinds` goes into the optional bucket and can get dropped once the budget runs out, with no `BudgetImpossible` raised.

So a policy file that sets `always_include_kinds` to something that leaves out `CONSTRAINT` (for example just `["DECISION"]`) will silently drop CONSTRAINT nodes under budget pressure instead of raising. Checked this directly:

```python
pol = ProjectorPolicy(always_include_kinds=frozenset({"DECISION"}), default_budget=10)
nodes = [{"id": "c1", "kind": "CONSTRAINT", "payload": {"rule": "must-never-drop"}, "step_n": 1, "seq": 0}]
result = project_context(nodes, policy=pol, budget=5)
# result.dropped_ids == ('c1',), no exception raised
```

README says the "never silently drop a CONSTRAINT" invariant has to hold even for a custom policy file, and R04 specifically says a CONSTRAINT node either gets included or the projector raises a hard error. `test/unit/test_projector_policy.py`, `test/unit/test_projector.py`, and `conformance/r04_constraint_budget_test.py` only exercise the default policy or policies that already include CONSTRAINT, so this gap currently has no test coverage.

## Fix

Force `CONSTRAINT` into the must-include set regardless of what a policy supplies, either by rejecting a policy that omits it in `__post_init__` / `policy_from_mapping`, or by unioning it in unconditionally inside `project_context`. Add a regression test with a policy that deliberately omits CONSTRAINT from `always_include_kinds` and assert `BudgetImpossible` still fires.

## Spec reference

README §9 (default projector policy invariant), conformance R04

jakharmonika364:
## Problem

ProjectorPolicy checked that always_include_kinds was non-empty but never checked that CONSTRAINT was actually in it. A policy built with, say, always_include_kinds=frozenset({"DECISION"}), whether constructed directly or loaded through policy_from_mapping from a file that lists other kinds and omits CONSTRAINT, gets accepted unchanged. project_context then silently drops CONSTRAINT nodes for that policy instead of raising BudgetImpossible.

## Fix

__post_init__ now adds CONSTRAINT to always_include_kinds whenever it's missing, using object.__setattr__ since the dataclass is frozen. This covers both direct construction and policy_from_mapping, since the latter always goes through the ProjectorPolicy constructor.

## Testing

Added test/unit/test_projector_policy_constraint_guard.py covering direct construction without CONSTRAINT, policy_from_mapping without CONSTRAINT, and the explicit-CONSTRAINT case staying unaffected. Full test/unit suite passes locally (99 passed, 2 skipped, unrelated to this change).

Closes #90

### DIFF
--- a/pkg/trajectory_ir/runtime/policy.py
+++ b/pkg/trajectory_ir/runtime/policy.py
@@ -51,6 +51,12 @@ class ProjectorPolicy:
             raise PolicyError("default_budget must be non-negative")
         if not self.always_include_kinds:
             raise PolicyError("always_include_kinds must be non-empty")
+        if "CONSTRAINT" not in self.always_include_kinds:
+            # CONSTRAINT must always survive projection (README §4); a
+            # policy that omits it is a footgun, not a real customization.
+            object.__setattr__(
+                self, "always_include_kinds", self.always_include_kinds | {"CONSTRAINT"}
+            )
 
 
 def default_projector_policy() -> ProjectorPolicy:
--- a/test/unit/test_projector_policy_constraint_guard.py
+++ b/test/unit/test_projector_policy_constraint_guard.py
@@ -1,0 +1,18 @@
+from trajectory_ir.runtime.policy import ProjectorPolicy, policy_from_mapping
+
+
+def test_direct_construction_without_constraint_still_includes_it():
+    policy = ProjectorPolicy(always_include_kinds=frozenset({"DECISION"}))
+    assert "CONSTRAINT" in policy.always_include_kinds
+    assert "DECISION" in policy.always_include_kinds
+
+
+def test_policy_from_mapping_without_constraint_still_includes_it():
+    policy = policy_from_mapping({"always_include_kinds": ["DECISION"]})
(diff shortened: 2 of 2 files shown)


Siddhartha-singh01 (approved):
Nice footgun fix. Forcing CONSTRAINT into always_include_kinds matches the projector rule and the tests cover both construction paths. Approving and merging.

Issue #67, opened 2026-08-05
spec: reconcile master README §3.1/§5 backend scope with Go's Temporal adapter


jakharmonika364:
## Problem

The master README (`README.md`, authoritative per §0) names the pluggable durable execution backend as **DBOS (Phase 1A default) or Restate** only (§3.1, §5), and explicitly lists "*a second durable execution backend adapter (Restate or otherwise) before the DBOS adapter is conformant*" as **out of scope for this phase** (§5). Per §0's own precedence rule, this document is meant to be correct if anything else conflicts with it, and §5 tells contributor agents to raise an issue rather than implement anything on the out-of-scope list.

However, `go/trajir/durable/temporal` already ships as the Go port's stated production target (issues #16, #24, both closed), a **third** backend not named in the spec at all - and no spec-version bump (§14) was made to reflect it. Issue #66 (Restate for Python) explicitly excludes touching this, so it remains unreconciled.

## Proposal

Pick one:
1. Amend README §3.1/§5 (with a version bump per §14) to formally admit Temporal as a recognized backend option, documenting why Go diverged from the DBOS-first posture, or
2. Treat the Go Temporal adapter as a temporary/local decision pending a real Python-parity backend choice, and say so explicitly in `go/README.md` and the master spec's scope table, or
3. If Temporal was chosen for Go-ecosystem reasons DBOS doesn't serve, document that rationale in §3.1 so it isn't read as spec drift by future contributors/auditors.

## Spec reference

- README §0 (precedence), §3.1 (backend table), §5 (Phase 1A scope / out-of-scope list), §14 (version bump process)
- `go/README.md` "Durable backend decision (issues #16, #24)"
- Issue #66 (explicitly excludes this reconciliation)

**Why this matters**: anyone evaluating this project for CNCF hosting (the README's own stated audience) will hit this table and Go's actual dependency and see a contradiction with no explanation on file.

jakharmonika364:
@Ayush-Patel-56 @Siddhartha-singh01 what's your views on this ?

Siddhartha-singh01:
Thanks for filing this @jakharmonika364  the inconsistency is real and worth resolving in the spec rather than only in go/ docs.

## My understanding

Master README is authoritative (§0). On durable backends it currently does two different things:

1. **Prior art (§3.1):** Temporal, Restate, and DBOS are all named as engines we must *not* reimplement. Trajectory IR sits on top of “one of these.”
[...]

PR #102, opened 2026-08-06
docs: label Memory and LocalSQLite as durable backend test fakes

### LINKED ISSUE #92: fix: Go trajir/durable implements its own crash-safe step memoization
## Problem

`go/trajir/durable/backend.go` (line 22-75), `memory.go`, and `sqlite.go` implement a first writer wins step memoization engine from scratch: `Step`/`Infer`/`Tool` look up a stored result, run the function if it's missing, and save the result, backed by either an in-process map or a `step_memos` SQLite table with `INSERT OR IGNORE`. Neither DBOS nor Restate shows up anywhere in `go/trajir/durable` or `go.mod`.

README §12.0 names DBOS as the Phase 1A default backend and Restate as an optional second adapter, both external dependencies the project consumes rather than reimplements. Rule 2 in §15.2 is explicit about this: never implement custom crash detection, retry, or lease/heartbeat logic outside the backend adapter, a PR that adds that kind of logic is a defect regardless of how small it looks.

`doc.go` in that package is upfront about the choice, it documents it as issue #16 and names Temporal as the eventual production target from issue #24. That's a real scope decision made in code comments rather than through the `SPEC-QUESTION` process rule 1 asks for when the spec is silent on something.

Practical effect: `crashagent`, `kill-mid-deploy`, and the Go R01/R02 conformance tests all run against this hand rolled backend through `durable.OpenLocal`. Right now the Go conformance suite is validating a bespoke memoization engine, not an integration with a real durable execution backend.

This is related to #67, which is about reconciling the README with the Temporal adapter specifically. This issue is about the base `trajir/durable` package itself, the hand rolled memoization exists independently of whichever engine eventually backs it, and predates the Temporal work.

## Proposal

Not proposing a specific code fix here, this needs a decision on backend strategy first. Either wire `trajir/durable` into a real DBOS or Restate client and keep `Memory`/`LocalSQLite` around explicitly as test fakes, or open a formal `SPEC-QUESTION` to scope Go's backend story separately from the Python side. Flagging for triage alongside #67 rather than proposing changes here.

## Spec reference

README §12.0, §15.2 rule 1, §15.2 rule 2. Related: #67

jakharmonika364:
## Problem

go/trajir/durable/backend.go, memory.go, and sqlite.go implement a first writer wins step memoization engine from scratch, backed by an in process map or a SQLite table. Neither DBOS nor Restate shows up anywhere in that package, but the package doc only described the coding default without saying plainly that Memory and LocalSQLite are test fakes rather than a real durable execution backend integration. crashagent, kill-mid-deploy, and the Go R01/R02 conformance tests all run against this hand rolled store through durable.OpenLocal, and that isn't obvious without reading the implementation.

This issue itself doesn't propose a specific code fix, it flags the backend strategy decision for triage alongside #67. Wiring trajir/durable into a real DBOS or Restate client is a separate, larger decision that shouldn't happen inside a documentation PR.

## Fix

Documentation only change. doc.go, memory.go, and sqlite.go now say explicitly that Memory and LocalSQLite are test fakes, not a production backend, and point at this issue for the open wiring question. No functional code changed.

## Testing

go build ./... and go test ./trajir/durable/... both pass locally with no behavior change.

Closes #92

### DIFF
--- a/go/trajir/durable/doc.go
+++ b/go/trajir/durable/doc.go
@@ -2,10 +2,14 @@
 //
 // Decision (issue #16)
 //
-//	Coding default: LocalSQLite (and Memory for pure unit tests).
-//	It memoizes step results by workflow id and step key so re-entry skips
-//	re-running model inference and tools. No separate Temporal or Restate
-//	server is required for Phase 1B Go development.
+//	Coding default: LocalSQLite (and Memory for pure unit tests). Both are
+//	test fakes: a hand rolled first writer wins memoization store, not an
+//	integration with a real durable execution engine. They memoize step
+//	results by workflow id and step key so re-entry skips re-running model
+//	inference and tools, which is enough for Phase 1B Go development and
+//	the current conformance suite, but neither should be treated as a
+//	production backend. See issue #92 for the open question on wiring
+//	trajir/durable into a real DBOS or Restate client, tracked alongside #67.
 //
 //	Production target: Temporal as the first full Go adapter
 //	(go/trajir/durable/temporal, issue #24). Restate remains a welcome
--- a/go/trajir/durable/memory.go
+++ b/go/trajir/durable/memory.go
@@ -5,7 +5,8 @@ import (
 	"sync"
 )
 
-// Memory is an in process Backend. Fine for unit tests. Does not survive process exit.
+// Memory is a test fake Backend, not a durable execution engine integration.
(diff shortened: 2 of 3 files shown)


Siddhartha-singh01 (approved):
Clear docs change. Calling out Memory and LocalSQLite as test fakes (not real durable backends) will save people from the wrong mental model. Approving and merging.

PR #103, opened 2026-08-06
fix: fail closed on artifact manifest entries missing content_hash

### LINKED ISSUE #93: fix: ensure_artifacts_in_cas silently skips manifest entries with missing content_hash
## Problem

`ensure_artifacts_in_cas` in `pkg/trajectory_ir/storage/artifacts.py` (line 62-84) does this:

```python
for entry in artifacts_manifest:
    raw = entry.get("content_hash")
    if raw is None:
        continue
```

An entry with no `content_hash` just gets skipped instead of failing. `rehydrate_artifacts` in `pkg/trajectory_ir/storage/rehydrate.py` (line 36-38) raises `CASError("artifact manifest entry missing content_hash")` for the exact same case, so the two functions disagree on what to do here.

`ensure_artifacts_in_cas` is the fail closed pre-check `export_tir(mode="thin", cas=...)` relies on to guarantee a thin package never claims an artifact the store can't actually serve (`pkg/trajectory_ir/package/tir.py`, line 278-290). `ArtifactRef` is a plain dataclass with no field validation, so you can build one with `content_hash=None` and pass it through export. Export succeeds silently instead of raising `CASNotFoundError` like it's supposed to, and the failure only shows up later, at rehydrate time, on a completely different code path.

## Fix

Make `ensure_artifacts_in_cas` raise the same way `rehydrate_artifacts` does when `content_hash` is missing from an entry, instead of skipping it.

## Spec reference

README §9 (thin package rehydration), §4 principle 4

jakharmonika364:
## Problem

ensure_artifacts_in_cas skipped any manifest entry whose content_hash was None instead of raising. rehydrate_artifacts raises CASError for the exact same shape of entry, so the two functions disagreed.

ensure_artifacts_in_cas is the fail closed pre-check export_tir(mode="thin", cas=...) relies on to guarantee a thin package never claims an artifact the store cannot actually serve. ArtifactRef is a plain dataclass with no field validation, so building one with content_hash=None and passing it through export succeeded silently instead of raising CASNotFoundError. The failure only showed up later, at rehydrate time, on a completely different code path.

## Fix

ensure_artifacts_in_cas now raises CASError the same way rehydrate_artifacts does when content_hash is missing from an entry, regardless of require_all, since a missing content_hash is a manifest shape problem rather than a "hash not found in store" case.

## Testing

Added test/unit/test_ensure_artifacts_fail_closed.py covering a missing key, an explicit None value, the require_all=False case, and the passing case with a real hash. Full test/unit suite passes locally (100 passed, 2 skipped, unrelated to this change).

Closes #93

### DIFF
--- a/pkg/trajectory_ir/storage/artifacts.py
+++ b/pkg/trajectory_ir/storage/artifacts.py
@@ -15,7 +15,13 @@ from __future__ import annotations
 from typing import Any
 
 from trajectory_ir.package.tir import ArtifactRef
-from trajectory_ir.storage.cas import CAS, CASNotFoundError, content_hash, normalize_content_hash
+from trajectory_ir.storage.cas import (
+    CAS,
+    CASError,
+    CASNotFoundError,
+    content_hash,
+    normalize_content_hash,
+)
 
 
 def put_artifact(
@@ -74,7 +80,7 @@ def ensure_artifacts_in_cas(
     for entry in artifacts_manifest:
         raw = entry.get("content_hash")
         if raw is None:
-            continue
+            raise CASError("artifact manifest entry missing content_hash")
         h = normalize_content_hash(str(raw))
         if not store.has(h):
             missing.append(h)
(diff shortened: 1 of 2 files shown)


Siddhartha-singh01 (approved):
Right call. Skipping missing content_hash was too soft for package integrity. Fail closed with CASError and the tests look solid. Approving and merging.

PR #105, opened 2026-08-06
fix: write .tir packages atomically instead of straight to dest_path

### LINKED ISSUE #95: fix: .tir package export is not atomic
## Problem

`export_tir` in `pkg/trajectory_ir/package/tir.py` writes straight to the final `dest_path` (line 321):

```python
with zipfile.ZipFile(dest_path, "w", compression=zipfile.ZIP_DEFLATED) as zf:
```

No temp file, no `os.replace`. The CAS layer in `pkg/trajectory_ir/storage/fs.py` (line 78-90) does this correctly: write to a temp file, fsync, then atomic rename, specifically so readers never observe a partial object. Package export doesn't follow the same pattern.

If the process gets killed mid export, `dest_path` ends up holding a truncated zip. A caller that just checks `dest_path.is_file()` after calling `export_tir` sees what looks like a successful export but is actually a broken package.

## Fix

Write to a temp path in the same directory as `dest_path`, then `os.replace` it into place once the zip is fully written and flushed, matching what `fs.py` already does for CAS objects.

## Spec reference

README §4 principle 4 (correctness should not depend on partial writes being safe to read)

Ayush-Patel-56:
## Problem

export_tir wrote the zip straight to dest_path with zipfile.ZipFile(dest_path, "w", ...), no temp file, no os.replace. The CAS layer in storage/fs.py already does this correctly: write to a temp file, fsync, then atomic rename, specifically so readers never observe a partial object. Package export didn't follow the same pattern.

If the process got killed mid export, dest_path ended up holding a truncated zip. A caller that just checks dest_path.is_file() after calling export_tir would see what looks like a successful export but is actually a broken package. Re-exporting over an existing good package on a failed run could also clobber it with a partial file.

## Fix

export_tir now writes into a temp file created with tempfile.mkstemp in the same directory as dest_path, fsyncs it, and only calls os.replace once the zip is fully written and flushed. On any failure the temp file is cleaned up and dest_path is left untouched, matching what fs.py already does for CAS objects.

## Testing

Added test/unit/test_tir_export_atomic_write.py: one test confirms no temp file is left behind after a normal export, one monkeypatches zipfile.ZipFile.writestr to raise mid write and confirms an existing dest_path is left byte for byte unchanged with no stray temp file, and one confirms the exported package still loads normally. Full test/unit suite passes locally (99 passed, 2 skipped, unrelated to this change).

Closes #95

### DIFF
--- a/pkg/trajectory_ir/package/tir.py
+++ b/pkg/trajectory_ir/package/tir.py
@@ -20,6 +20,8 @@ from __future__ import annotations
 
 import hashlib
 import json
+import os
+import tempfile
 import zipfile
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Siddhartha-singh01 (approved):
Solid reliability fix. Temp file + fsync + replace means a crash mid export cannot leave a half written package at dest. Failure path and tests look good. Approving and merging.

PR #106, opened 2026-08-06
fix: propagate real errors from claim_tool_call instead of swallowing them

### LINKED ISSUE #96: fix: Postgres claim_tool_call treats any exception as a lost race
## Problem

`claim_tool_call` in `drivers/postgres/log.py` (line 168-206) wraps the whole SELECT FOR UPDATE / INSERT / commit block in a bare `except Exception` (line 203). On any failure it rolls back and returns `False`, which the caller reads as "someone else already claimed this slot." That's correct behavior for an actual unique violation, but the same except also swallows connection drops, deadlocks, and anything else that can go wrong in that block.

If the DB connection drops mid claim, the caller gets `False`, assumes another worker already claimed and is executing the non-idempotent tool, and never retries or surfaces an error. The step gets silently skipped instead of failing loudly, which is the opposite of what the block-and-gate policy is supposed to guarantee for non-idempotent calls.

## Fix

Narrow the except to the actual unique-violation exception type (`psycopg.errors.UniqueViolation` or equivalent) and let anything else propagate as a real error instead of being interpreted as a lost race.

## Spec reference

README §8.3 (block and gate should fail loud, not silently skip a non-idempotent call)

Ayush-Patel-56:
## Problem

`PostgresNodeLog.claim_tool_call` wrapped its SELECT-then-INSERT claim attempt in a bare `except Exception: rollback(); return False`. That means a genuine slot conflict (another worker already claimed the `TOOL_CALL` slot) and a real failure such as a dropped connection or a deadlock both produce the exact same observable result: `False`.

A caller receiving `False` reasonably concludes "another worker already claimed this, skip it." If the `False` actually came from a dropped connection mid-claim, the step gets silently skipped instead of the failure being surfaced, which is the opposite of the fail-closed behavior the rest of the node log relies on.

## Fix

Narrow the except clause in `drivers/postgres/log.py`'s `claim_tool_call` to catch only `psycopg.errors.UniqueViolation`, the actual signal for "lost the race" (mirrors the local-import pattern already used in `append()`). Any other exception now rolls back and re-raises, so real failures propagate instead of being reinterpreted as a lost claim.

## Testing

Added `test/unit/test_postgres_claim_tool_call_exception_narrowing.py` with minimal, self-contained fake connection/cursor objects (does not touch the existing `_FakeStore`/`_FakeCursor` in `test_postgres_node_log.py`):
- `test_claim_tool_call_returns_false_on_unique_violation`: a genuine `UniqueViolation` on the INSERT still returns `False`, preserving existing "lost the race" semantics.
- `test_claim_tool_call_propagates_unrelated_exception`: an unrelated exception (simulating a dropped connection) now propagates instead of being swallowed as `False`.

Full `test/unit` suite passes (98 passed, 2 skipped — the skips are pre-existing, gated on live Postgres/pip-audit).

Closes #96

---
This fix was developed with AI assistance (Claude Code) and reviewed by the author before submission.

### DIFF
--- a/drivers/postgres/log.py
+++ b/drivers/postgres/log.py
@@ -206,6 +206,9 @@ class PostgresNodeLog:
         seq: int,
     ) -> bool:
         """Atomically claim a TOOL_CALL slot (single winner under concurrency)."""
+        # Import here so the module loads without psycopg installed.
+        from psycopg.errors import UniqueViolation
+
         node = Node(
             kind="TOOL_CALL",
             trajectory_id=trajectory_id,
@@ -249,10 +252,13 @@ class PostgresNodeLog:
                     )
                 self._conn.commit()
                 return True
+            except UniqueViolation:
+                self._conn.rollback()
+                # Another writer won the slot between our SELECT and INSERT.
+                return False
             except Exception:
                 self._conn.rollback()
-                # Unique violation: another writer won the slot.
-                return False
+                raise
 
     def has(self, trajectory_id: str, step_n: int, kind: str, seq: int | None = None) -> bool:
         sql = "SELECT 1 FROM nodes WHERE trajectory_id = %s AND step_n = %s AND kind = %s"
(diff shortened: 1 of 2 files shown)


Siddhartha-singh01 (approved):
Important fail closed fix. UniqueViolation should mean lost race; a dropped connection should not look the same. Matches the pattern already used in append(). Approving and merging.

PR #107, opened 2026-08-06
fix: make resume() a tested, documented reattach instead of a silent stub

### LINKED ISSUE #97: fix: client SDK resume() does not actually resume anything
## Problem

`resume` in `client/python/trajectory_client.py` (line 139-152) is one of the six functions README §12.1 names as the client SDK surface: `open_trajectory`, `project`, `seal_decision`, `exec_tool`, `commit_step`, `resume`. Its body just calls `init_backend` and constructs the same `Trajectory` dataclass `open_trajectory()` does. It doesn't inspect any prior DECISION/TOOL_RESULT state and doesn't do any replay of its own, it's functionally the same function under a different name.

Grepped the repo, nothing calls `client.python.trajectory_client.resume` anywhere except its own definition. The actual resume behavior that exists in this codebase, `examples/kill-mid-deploy/agent.py`, works by re-invoking `resume/step.py`'s machinery directly under the same durable workflow id, bypassing the client SDK's `resume()` entirely.

So one of the six named SDK functions is effectively a stub that happens to work by accident, since the real resume logic lives one layer down and doesn't route through it, untested, and unused by the one example built specifically to demonstrate resume.

## Fix

Either make `resume()` do something real, wire it to the same step replay path `kill-mid-deploy` uses so a caller going through the public SDK actually gets resume semantics, or if `open_trajectory()` genuinely is sufficient for resuming since the durable backend keys off workflow id either way, document `resume()` as an explicit alias and add a test that exercises it through an actual crash and restart scenario instead of leaving it uncovered.

## Spec reference

README §12.1

Ayush-Patel-56:
## Problem

`resume()` in `client/python/trajectory_client.py` (one of the six SDK functions named in README §12.1) was byte-for-byte identical to `open_trajectory()`: it never inspected any prior `DECISION`/`TOOL_RESULT` state and did no replay of its own. Nothing in the codebase calls it. The one example built specifically to demonstrate resume, `examples/kill-mid-deploy/agent.py`, bypasses the client SDK entirely and re-invokes `resume/step.py`'s `make_run_step` machinery directly under the same durable workflow id. So `resume()` looked like a real SDK entry point but was an untested, unused stub that happened to work by accident.

## Fix

This SDK layer does not wrap steps in a durable workflow decorator the way `make_run_step` does. Its crash safety comes entirely from `NodeLog` being idempotent by content and from `exec_tool`'s block-and-gate claim on `NON_IDEMPOTENT_WRITE` tools (README §8, R02). Re-driving the same `step_n`/`seq` calls against the same `db_path` is what "resume" means at this layer, so there genuinely is no separate replay step to wire in.

Rather than leave that implicit, `resume()`:
- Now has a docstring explaining why it's a thin wrapper and what actually makes resumption safe.
- Raises `ValueError` if the trajectory has no existing nodes in the log, since resuming something with no history is almost always a caller bug (wrong `db_path` or `trajectory_id`) rather than a legitimate resume.

## Testing

Added `test/unit/test_client_sdk_resume_alias.py`:
- `test_resume_raises_on_trajectory_with_no_history`: resuming a trajectory with an empty log raises `ValueError`.
- `test_resume_reattaches_after_history_exists`: resuming after some history exists returns a `Trajectory` with the correct fields.
- `test_resume_does_not_replay_already_claimed_gated_tool_call`: an actual crash/restart scenario, mirroring `kill-mid-deploy`'s approach. Execs a gated `NON_IDEMPOTENT_WRITE` tool, drops the in-process `Trajectory`, calls `resume()`, then re-drives the same `exec_tool` call and asserts it blocks with `BlockedNeedsGate` instead of re-running the tool body.

Full `test/unit` suite passes (99 passed, 2 skipped — pre-existing skips gated on live Postgres/pip-audit).

Closes #97

---
This fix was developed with AI assistance (Claude Code) and reviewed by the author before submission.

### DIFF
--- a/client/python/trajectory_client.py
+++ b/client/python/trajectory_client.py
@@ -143,7 +143,27 @@ def resume(
     *,
     mode: RunMode | str = RunMode.LIVE,
 ) -> Trajectory:
+    """Reattach to a trajectory that already has node log history.
+
+    This SDK does not wrap steps in a durable workflow decorator the way
+    ``trajectory_ir.resume.step.make_run_step`` does; crash safety here comes
+    entirely from ``NodeLog`` being idempotent by content and from
+    ``exec_tool``'s block-and-gate claim on NON_IDEMPOTENT_WRITE tools
+    (README S8, R02). Re-driving the same step_n/seq calls against the same
+    db_path is what "resume" means at this layer, so there is no separate
+    replay step to run here beyond reconnecting to the log.
+
+    Unlike ``open_trajectory``, this raises if ``trajectory_id`` has no prior
+    nodes: resuming a trajectory with nothing to resume is almost always a
+    caller bug (wrong db_path or trajectory_id), and should fail loudly
+    rather than silently behave like ``open_trajectory``.
+    """
     init_backend(app_name=trajectory_id)
+    if not NodeLog(db_path).list_nodes_all_tenants(trajectory_id):
+        raise ValueError(
+            f"cannot resume trajectory_id={trajectory_id!r}: no existing nodes "
+            f"found in {db_path!r}; use open_trajectory() to start a new one"
+        )
     return Trajectory(
         trajectory_id=trajectory_id,
         tenant_id=tenant_id,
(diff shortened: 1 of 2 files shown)


Siddhartha-singh01 (approved):
Honest and useful. resume() now fails loudly when there is no history, and the crash style test for gated tools is exactly the behavior we want. Approving; may need a quick rebase after #100 lands since both touch trajectory_client.py.

Issue #128, opened 2026-08-07
ci: unlock main gates, split fast vs deep CI, Go first for Phase 1B


Siddhartha-singh01:
## Problem

CI on Trajectory IR is already fairly complete (DCO, Python quality matrix, package smoke, pip-audit, Go tests + govulncheck, Postgres/MinIO integration). Three gaps keep it from being a strong CI/CD spine for Phase 1B:

1. **Main is not really gated.** The repo is private on a free plan, so branch protection / rulesets return 403. Green checks are process only: a red PR can still be merged if someone has write access.
2. **One flat CI surface.** Every PR pays for everything (or nothing is ordered). There is no clear **fast gate** (must be green in a few minutes) vs **deep gate** (heavier, still required, can be slower). That makes it harder to add Go integration and raise floors without slowing every contributor to a crawl.
3. **Go CI is still secondary.** Python runs ruff, mypy, 80% coverage, e2e, and full R01 through R08. Go runs unit coverage at a **50%** floor plus govulncheck. For Phase 1B (Go primary SDK, epic #113) that priority is backwards: new Go drivers and demos need the same or stronger enforcement than the Python reference port.

This issue tracks **all three** workstreams in one place so we do not open three thin issues that get half done.

### 1) Unlock real gates on main (highest impact)

**Intent:** Merges into `main` must require green CI, not goodwill.

**Do:**

1. Choose how protection becomes available:
   - make the repository **public**, or
   - upgrade org/user plan so **private** repos get branch protection / rulesets.
2. Protect `main`:
   - require pull request before merge
   - require status checks (exact Action check names)
   - **require branch up to date with main**
   - block force push and branch deletion
   - require at least one approval when more than one maintainer is active
3. Required check names (start with current stable set, then expand):
   - `DCO`
   - `Quality (Python 3.11)`
   - `Quality (Python 3.12)`
   - `Package smoke`
   - `Security (pip-audit)`
   - `Go`
   - then, once green for a stretch: `Integration (Postgres)`, `Integration (MinIO)`
4. Enable secret scanning / push protection when the plan allows; keep Dependabot alerts + security updates on.
5. Update `docs/maintainer-branch-protection.md` with the final required check list and up to date note.

**Acceptance (part 1):**

1. Branch protection or rulesets active on `main` (no 403).
2. A PR with a failing required check cannot merge.
3. Docs match the live required check names.

### 2) Split fast gates vs deep gates

[...]

PR #139, opened 2026-08-08
test: gate CI on Python to Go .tir cross language round trip

### LINKED ISSUE #130: test: Python and Go .tir round trip in CI
## Parent

Phase 1B milestone. Epic #113.

## Problem

Hash goldens already cross check Python and Go. Portable packages need the same: a thin (and ideally fat) `.tir` written by one language must load and verify in the other.

## Proposal

1. CI step or small harness: Python `export_tir` fixture then Go `tir.Load` / import; reverse if cheap.
2. Fail closed on node id or seal mismatch.
3. Use existing testdata or generate in job temp dir.
4. Place under deep gate once #128 job split exists; until then a dedicated step in Quality or Go job is fine.

## Acceptance

1. Automated test on CI proves at least thin Python to Go round trip (and document Go to Python if implemented).
2. Uses real package APIs, no invented format.
3. Green on main.

## Out of scope

Package signatures, multi tenant SaaS.

Ayush-Patel-56:
## Summary

Issue #130 asks for a CI gate proving a thin `.tir` package written by Python loads and verifies in Go. Most of the machinery already existed and just wasn't wired into CI as a real gate:

- `scripts/gen_tir_fixture.py` already builds a minimal trajectory and calls the real `export_tir` API to write `testdata/sample_thin.tir`.
- `go/trajir/tir/tir_test.go` already had `TestImportPythonGoldenFixture`, which calls the real `tir.Import` API against that fixture and asserts node counts match. `tir.Import` unconditionally verifies node ids and seal `content_hash`, so a mismatch already fails closed (see `TestImportDetectsTamperedNode` in the same file for proof that tampering is caught).

Two problems: that test only ran incidentally inside the `go test ./...` catch-all step, and it silently `t.Skip`s if the fixture is missing rather than failing, so a deleted or never-generated fixture would not actually gate anything. And the committed fixture was static, so it would not catch a future Python-side `export_tir` change that broke Go-side compatibility.

This PR:

1. Changes `TestImportPythonGoldenFixture`'s missing-fixture branch from `t.Skipf` to `t.Fatalf`, so a missing fixture is a hard CI failure, not a silent skip.
2. Adds two explicit steps to the `Go` job in `.github/workflows/ci.yml`: `python scripts/gen_tir_fixture.py` (regenerates the fixture fresh from current Python `export_tir` on every run, using only the stdlib, no extra installs needed) followed by `go test ./trajir/tir -run TestImportPythonGoldenFixture -count=1 -v` as its own named step, so the round trip is a visible, dedicated gate rather than buried in the full suite run.
3. Documents the check in `CONTRIBUTING.md`, including that the reverse direction (Go `tir.Export` read back by Python `load_tir`) is not wired up yet.

Per #128's still-open part 2 (fast/deep split, not yet merged), this lands as a step in the existing `Go` job for now, matching the issue's own guidance ("until then a dedicated step in Quality or Go job is fine").

## Related issues

Closes #130 for the thin Python to Go direction (the acceptance criteria's minimum bar). Go to Python is called out as a documented follow up, not implemented here, since it needs a new Go-side fixture-writer entry point that does not exist yet.

## How I tested

Locally: deleted `testdata/sample_thin.tir`, ran `python scripts/gen_tir_fixture.py` to regenerate it, then `go test ./trajir/tir -run TestImportPythonGoldenFixture -count=1 -v` (pass). Also ran `go vet ./...`, `go build ./...`, and the full `go test ./trajir/tir/...` package suite (pass). Validated the workflow YAML parses. Left the committed `testdata/sample_thin.tir` unchanged since CI regenerates it fresh (zip output isn't byte stable across regenerations, but both stay semantically valid).

## Checklist

- [x] Change matches the master README (no invented APIs, reuses `export_tir` / `tir.Import`)
- [x] Relevant CI checks pass locally
- [ ] AI assistance disclosed below


jakharmonika364 (approved):
lgtm

Issue #131, opened 2026-08-08
ci: raise Go trajir coverage floor toward 70 then 80


Siddhartha-singh01:
## Parent

Phase 1B milestone. Epic #113. Related: #128 part 3.

## Problem

`GO_COV_FAIL_UNDER` is 50 while Python unit floor is 80. Go is Phase 1B primary; a 50 percent floor is too soft for merge gates.

## Proposal

1. Measure current total on `go/trajir/...` after #124/#125 land.
2. Raise floor in two steps (for example 50 to 70, later 70 to 80) only after green baseline.
3. Prefer tests on low coverage packages (graft, redact, sandbox, new drivers) over gaming the metric.
4. Document floors in maintainer branch protection notes.

## Acceptance

1. CI env floor is at least 70 with green main, with a written path to 80.
2. No floor drop without a PR.
3. Coverage script still enforces the floor.

## Out of scope

Python coverage changes (separate if needed).

PR #140, opened 2026-08-08
test: raise Go trajir coverage floor from 50 to 70 (#131)

Ayush-Patel-56:
## Summary

Issue #131 asks to raise `GO_COV_FAIL_UNDER` toward 70 then 80. Measured baseline on `main` before this PR was 65.0% statements on `go/trajir/...`, below the 70 acceptance bar. Rather than just moving the number (which would fail CI on `main`), this PR adds real unit tests to close the gap honestly:

- `graft`: `Error.Error()`, `FindArtifactRef` empty/no-match/ordering cases, `GraftArtifactRef` nil target and explicit-override paths.
- `nodes`: `ValidateIDComponent` rejection cases (empty, invalid UTF-8, delimiter, control char, too long), `PayloadHash` marshal-error path, `NodeID`/`NewNode` invalid-component paths.
- `redact`: non-string leaf pass-through, list redaction, non-map payload, nil context, free-form dict context, non-map item skipping.
- `resume`: gated tool error propagation, `RunStep` nil Log/Backend, model error, missing/unnamed `tool_calls`, unknown tool, sandbox-mode block, non-gated tool error.
- `sandbox`: `NormalizeMode` table (empty/live/sandbox/case-insensitive/invalid), `Forbidden.Error()`.
- `tir`: `Export`/`Import`/`Load` error paths (nil `NodeLog`, unsupported mode, unknown trajectory, mixed tenants, fat-mode missing/mismatched/oversized artifact bytes, missing zip member, malformed manifest, unexpected artifact path).

Measured total after these tests: **70.2%** statements on `go/trajir/...`. `GO_COV_FAIL_UNDER` is raised from 50 to 70 to match.

`postgres` (4.9%) and `durable/temporal` (28.7%) are the packages holding the ceiling down: their real code only runs against a live Postgres/Temporal service in the separate Integration jobs, not in the `Go` job this floor gates. Getting to 80 needs either mocked coverage for those two packages or computing the floor with them excluded, both left as explicit follow-up (documented in `docs/maintainer-branch-protection.md`) rather than attempted here, since that's a bigger design decision than this PR's scope.

## Related issues

Addresses #131's step 1 (raise floor toward 70, in two steps 50 to 70 later 70 to 80). Floor is at 70 with a measured green baseline of 70.2%; the path to 80 is written up but not implemented in this PR.

## How I tested

`go vet ./...`, `go build ./...`, and `go test ./...` all clean. Ran `scripts/check_go_coverage.sh` locally with `GO_COV_FAIL_UNDER=70` (passes at 70.2%) and with the prior default (also passes). Validated the workflow YAML parses. All new tests pass individually and as part of the full suite.

## Checklist

- [x] Change matches the master README (no invented APIs; tests exercise existing exported surfaces only)
- [x] Relevant CI checks pass locally
- [ ] AI assistance disclosed below


jakharmonika364 (approved):
lgtm

PR #213, opened 2026-08-14
fix: rename kill-mid-deploy to kill_mid_deploy to fix invalid python namespace (#203)

### LINKED ISSUE #203: bug: Missing `__init__.py` files in `examples/` subdirectories cause namespace collisions
### What happened

The root cause of Issue 1 is that subdirectories under `examples/` (like `kill-mid-deploy` and `adoption_host`) lack `__init__.py` files. Python treats them as loose scripts rather than a package, leading to conflicting module names when static analyzers scan the tree.

### What you expected

Each subdirectory inside `examples/` should contain an `__init__.py` to properly namespace the example scripts.

### Steps to reproduce

1. Observe the directory structure in `examples/`
2. Note the absence of `__init__.py` in subdirectories.

### Environment

Any

### Logs or output

```text
N/A
```

N/A

desireddymohithreddy0925:
Closes #203

**Description**
The `__init__.py` files in `examples/kill-mid-deploy/` were being ignored by Python's module resolution because hyphens (`-`) are invalid characters in Python package names. This caused the folder to be treated as a collection of loose scripts instead of a proper package, triggering the namespace collisions.

**Changes**
- Renamed `examples/kill-mid-deploy/` to `examples/kill_mid_deploy/`.
- Updated all references to the folder path in `pyproject.toml`, Markdown documentation, and test fixture imports to use the new `kill_mid_deploy` naming convention.

**Verification**
- The `examples/kill_mid_deploy` directory is now correctly treated as a valid Python package namespace by static analyzers.

### DIFF
--- a/examples/kill_mid_deploy/agent.py
+++ b/examples/kill_mid_deploy/agent.py
@@ -1,14 +1,14 @@
 """Kill-mid-deploy fixture agent.
 
 Runs one durable step: infer a plan, seal the decision, execute a fake
-`deploy_server` tool. Used by test/e2e, conformance/, and the kill-mid-deploy
+`deploy_server` tool. Used by test/e2e, conformance/, and the kill_mid_deploy
 demo -- one script, three consumers, so crash-recovery behavior is only
 implemented once.
 
 Run it twice against the same working directory to exercise crash recovery:
 
-    python examples/kill-mid-deploy/agent.py --crash-during=tool_call   # then kill -9
-    python examples/kill-mid-deploy/agent.py --resume
+    python examples/kill_mid_deploy/agent.py --crash-during=tool_call   # then kill -9
+    python examples/kill_mid_deploy/agent.py --resume
 
 Crash points, each idling at a marker file so a harness can hard-kill there:
 `--crash-during=inference` (before the decision is sealed),
@@ -25,7 +25,7 @@ import os
 import sys
 import time
 
-# Run directly from a checkout (`python examples/kill-mid-deploy/agent.py`)
+# Run directly from a checkout (`python examples/kill_mid_deploy/agent.py`)
 # without requiring an installed package or a preset PYTHONPATH.
 _REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 for _path in (_REPO_ROOT, os.path.join(_REPO_ROOT, "pkg")):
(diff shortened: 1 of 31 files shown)


jakharmonika364:
refer https://github.com/Coder-s-OG-s/Trajectory-IR/pull/212#pullrequestreview-4942969091

Ayush-Patel-56 (commented):
can u please sign off DCO ?

desireddymohithreddy0925:
> can u please sign off DCO ?

sure bro @Ayush-Patel-56

Ayush-Patel-56:
please sign off DCO

examples/kill_mid_deploy/agent.py
```diff
@@ -147,7 +147,7 @@ def seal_marker_hook():
         on_decision_sealed=seal_marker_hook,
     )
 
-    init_backend(app_name="kill-mid-deploy")
+    init_backend(app_name="kill_mid_deploy")
```
Ayush-Patel-56, line 150: `app_name` now equals `DB_PATH`'s stem, so the DBOS system db and the NodeLog db both resolve to `kill_mid_deploy.sqlite`. Please use a distinct `app_name` (e.g. `kill_mid_deploy_dbos`) so they don't collide.

test/e2e/_util.py
```diff
@@ -17,7 +17,7 @@
     "tool_started.marker",
     "inference_started.marker",
     "kill_mid_deploy.sqlite",
-    "kill-mid-deploy.sqlite",
+    "kill_mid_deploy.sqlite",
```
Ayush-Patel-56, line 20: This duplicate is a symptom of the `app_name` collision in `agent.py`, should resolve once that's fixed, but please double check the two entries point to two real, distinct files.

Ayush-Patel-56 (changes requested):
This PR renames all doc references to `./examples/kill_mid_deploy`, but `go/examples/kill-mid-deploy/` was never renamed on disk, `go build ./examples/kill_mid_deploy/...` fails. Please rename the directory (or its `-workdir` default / README refs) so they match.

Siddhartha-singh01 (changes requested):
## Review (request changes)

Renaming Python `examples/kill-mid-deploy` → `kill_mid_deploy` is the right fix for invalid namespace / tooling issues (#203). Scope has grown past that.

### Blockers (also called out by Ayush)
1. **Go path inconsistency:** docs/CHANGELOG say `go/examples/kill_mid_deploy`, but on-disk Go demo is still `go/examples/kill-mid-deploy/`. Either rename the Go directory too, or keep hyphen only for Go and stop rewriting Go docs to underscores. Today `go build ./examples/kill_mid_deploy/...` fails.
2. **Do not add `issues.md`**  that looks like a dump of issue text into the tree; not part of the rename. Please drop it.
3. Large churn in `docs/superpowers/**` historical plans is unrelated noise for this bugfix  prefer leave history alone unless required.

### Also
- Empty `__init__.py` additions are fine **after** the Python rename is consistent.
- Rebase on latest `main` (several quality PRs landed since).

### Verdict
**Request changes** tighten to rename + reference updates only, keep Go/Python path story consistent, drop `issues.md`.

PR #214, opened 2026-08-14
fix: configure pytest pythonpath to resolve module imports (#204)

### LINKED ISSUE #204: bug: PyTest collection fails out-of-the-box due to missing PYTHONPATH resolution
### What happened

Running `pytest conformance/ -v` directly on a fresh clone (even after installing dependencies) fails with `ModuleNotFoundError: No module named 'trajectory_ir'` because the project uses a flat layout (`pkg/trajectory_ir`) without dynamically injecting `pkg` into the PYTHONPATH.

### What you expected

Tests should be discoverable and executable immediately, or `pyproject.toml` should configure pytest to include `pkg` in the python path (e.g., via `pythonpath = "pkg"` in `[tool.pytest.ini_options]`).

### Steps to reproduce

1. `python3 -m pip install pytest`
2. `pytest conformance/ -v`

### Environment

macOS, Python 3.14

### Logs or output

```text
ModuleNotFoundError: No module named 'trajectory_ir'
```

pyproject.toml / tool.pytest.ini_options

desireddymohithreddy0925:
Closes #204

**Description**
Pytest collection was failing completely out-of-the-box (`ModuleNotFoundError: No module named 'trajectory_ir'`) because the project uses a flat `pkg/` layout and depends on `hatchling` to build the package, which means `pytest` didn't inherently know where the source code was located without explicit path resolution.

**Changes**
- Added `pythonpath = ["pkg", "."]` to the `[tool.pytest.ini_options]` block in `pyproject.toml` so that Pytest dynamically injects the correct source directories into `sys.path` during execution and collection.

**Verification**
- Test collection now successfully resolves imports for both `trajectory_ir` (in `pkg/`) and `drivers/` without crashing instantly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ packages = ["pkg/trajectory_ir", "drivers", "client"]
 
 [tool.pytest.ini_options]
 testpaths = ["test", "conformance"]
+pythonpath = ["pkg", "."]
 addopts = [
     "-ra",
     "--strict-markers",


jakharmonika364 (changes requested):
refer https://github.com/Coder-s-OG-s/Trajectory-IR/pull/212#pullrequestreview-4942969091

Ayush-Patel-56 (commented):
can u please sign off DCO ?

jakharmonika364 (approved):
LGTM, thanks!

PR #220, opened 2026-08-15
fix: remove brittle sys.path injection from examples (#208)

### LINKED ISSUE #208: bug: Dynamic `sys.path` injection in `examples/kill-mid-deploy/agent.py` breaks tooling
### What happened

According to comments and Ruff configuration in `pyproject.toml`, `examples/kill-mid-deploy/agent.py` dynamically alters `sys.path` to allow local imports. This is a brittle anti-pattern that breaks static analysis, IDE autocomplete, and linting rules (forcing `E402` to be disabled for that file).

### What you expected

The examples should rely on proper package installation (`pip install -e .`) rather than hacking `sys.path` at runtime.

### Steps to reproduce

1. Open `pyproject.toml` and look at `[tool.ruff.lint.per-file-ignores]`.
2. See `"examples/kill-mid-deploy/agent.py" = ["E402"]` and the related comment.

### Environment

Any

### Logs or output

```text
N/A
```

pyproject.toml

desireddymohithreddy0925:
Closes #208

**Description**
The example scripts (`agent.py` and `run_host.py`) were dynamically altering `sys.path` at runtime to allow local module imports without an installation. This is a brittle anti-pattern that breaks static analysis, IDE autocomplete, and forces us to bypass linting rules like `E402`.

**Changes**
- Removed the `sys.path.insert(0, _path)` blocks from `examples/kill-mid-deploy/agent.py` and `examples/host_loop/run_host.py`.
- Removed the per-file `E402` linting bypasses from `pyproject.toml`.
- Ran `ruff check --fix` to correctly sort the imports according to the new structure.

**Verification**
- Running `ruff check .` passes 100% cleanly without the `E402` overrides. Examples now properly expect the repository to be installed via `pip install -e .` prior to execution.

### DIFF
--- a/examples/adoption_host/run_demo.py
+++ b/examples/adoption_host/run_demo.py
@@ -29,11 +29,6 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
-# Allow ``python examples/adoption_host/run_demo.py`` without a prior editable install.
-_REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-for _path in (_REPO_ROOT, os.path.join(_REPO_ROOT, "pkg")):
-    if _path not in sys.path:
-        sys.path.insert(0, _path)
 
 from client.python.trajectory_client import (  # noqa: E402
     commit_step,
--- a/examples/host_loop/run_host.py
+++ b/examples/host_loop/run_host.py
@@ -16,12 +16,6 @@ import os
 import sys
 import tempfile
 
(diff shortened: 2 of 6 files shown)


jakharmonika364 (changes requested):
Now that this `sys.path` fallback is removed, `kill-mid-deploy/README.md`'s quickstart (`python examples/kill-mid-deploy/run_demo.py`) breaks on a fresh clone, it never mentions `pip install -e .`. Same gap in this file's docstring, and `examples/adoption_host/run_demo.py` still has the old hack this PR should've also cleaned up.

PR #221, opened 2026-08-15
fix: make rfc8785 import graceful to avoid crashing pytest collection (#209)

### LINKED ISSUE #209: bug: PyTest collection fails when `rfc8785` is missing, even in dry-runs
### What happened

### What you expected

### Steps to reproduce

### Environment

macOS, Python 3.14

### Logs or output

```text
    import rfc8785
```

Test Suite Architecture

desireddymohithreddy0925:
Closes #209

**Description**
When running PyTest in environments isolating specific driver tests or verifying basic syntax, the entire test suite collection would crash completely (`ModuleNotFoundError`) if the `rfc8785` core dependency was missing, because it was imported unconditionally at the root module level in `nodes.py`.

**Changes**
- Wrapped `import rfc8785` in a `try/except ImportError` block in `pkg/trajectory_ir/runtime/nodes.py`.
- Moved the `ModuleNotFoundError` to trigger only when `payload_hash()` is actually invoked at runtime.

**Verification**
- Merely collecting tests with PyTest no longer crashes the test runner on missing core dependencies.
- Tests that explicitly rely on payload hashing will correctly fail or be skipped individually if the dependency is absent.

### DIFF
--- a/pkg/trajectory_ir/runtime/nodes.py
+++ b/pkg/trajectory_ir/runtime/nodes.py
@@ -2,7 +2,10 @@ import hashlib
 import time
 from dataclasses import dataclass, field
 
-import rfc8785
+try:
+    import rfc8785  # type: ignore[import-not-found]
+except ImportError:
+    rfc8785 = None
 
 NODE_KINDS = frozenset(
     {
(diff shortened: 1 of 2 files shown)


jakharmonika364:
CI is failing, please fix it

PR #243, opened 2026-08-17
Fix: move dbos to optional dependencies

### LINKED ISSUE #206: bug: DBOS is forced as a core dependency instead of an optional backend
### What happened

In `pyproject.toml`, `dbos` is listed under the core `dependencies` array. Since Trajectory IR is meant to be a portable intermediate representation, forcing DBOS (a specific durable execution backend) on all installations bloats the footprint for users who only want to parse or serialize TIR packages. 

### What you expected

DBOS should be moved to `[project.optional-dependencies]` (e.g., `dbos = ["dbos"]`), similar to how `postgres` and `s3` are structured.

### Steps to reproduce

1. `pip install .`
2. Observe DBOS and its heavy dependencies being installed by default.

### Environment

Any

### Logs or output

```text
N/A
```

pyproject.toml

desireddymohithreddy0925:
Resolves #206 by moving dbos from core dependencies to optional-dependencies to avoid bloating the installation footprint for users not using the DBOS backend.
Fixes #206.

This PR moves `dbos` from the core `dependencies` to `[project.optional-dependencies]` in `pyproject.toml`. Since Trajectory IR is designed to be a portable intermediate representation, this change ensures that the DBOS durable execution backend and its heavy dependencies are not forcefully installed for users who only need to parse or serialize TIR packages. 

Users can now opt into it by installing via `pip install .[dbos]`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,14 +27,14 @@ classifiers = [
     "Topic :: Software Development :: Libraries",
 ]
 dependencies = [
-    "dbos",
     "rfc8785",
 ]
 
 [project.optional-dependencies]
-dev = ["pytest", "pytest-cov", "ruff", "mypy", "pip-audit"]
+dev = ["dbos", "pytest", "pytest-cov", "ruff", "mypy", "pip-audit"]
 postgres = ["psycopg[binary]>=3.1"]
 s3 = ["boto3"]
+dbos = ["dbos"]
 
 [project.urls]
 Homepage = "https://github.com/Coder-s-OG-s/Trajectory-IR"


Ayush-Patel-56:
CI failure, can u please fix it

jakharmonika364 (changes requested):
Good direction for #206, but heads up: `_resolve_durable_hooks` in `pkg/trajectory_ir/resume/step.py` defaults to the DBOS adapter whenever `make_run_step()` is called without explicit hooks, and `drivers/durable_backend/dbos/adapter.py` still does a hard top-level `from dbos import DBOS, DBOSConfig`. So a bare `pip install trajectory-ir` + default `make_run_step()` call will now `ModuleNotFoundError` instead of working, CI doesn't catch it because `.[dev]` also pulls in `dbos` here.

Can we lazy-import `dbos` inside the adapter functions instead, same pattern as `_require_psycopg` in `drivers/postgres/log.py`? Something like:

```
def _require_dbos():
    try:
        import dbos
    except ImportError as exc:
        raise ImportError('dbos not installed; pip install "trajectory-ir[dbos]"') from exc
    return dbos
```
and use `_require_dbos().DBOS...` in each function. Also need a `dbos = ["dbos"]` extra in `pyproject.toml`. That way the light install actually works and the error is clear when it doesn't.

Siddhartha-singh01 (changes requested):
## Review (request changes)

Direction matches #206 (DBOS should not bloat a parse/serialize-only install). Monica’s earlier review is correct and still applies.

### Blockers
1. **`pyproject.toml` alone is not enough.** `drivers/durable_backend/dbos/adapter.py` still does a hard top-level `import dbos`, and `make_run_step()` defaults into that adapter. After this PR, bare `pip install trajectory-ir` + default run_step will `ModuleNotFoundError`. CI stays green because `.[dev]` still installs dbos.
2. Please **lazy-import** DBOS (same pattern as `_require_psycopg` in `drivers/postgres/log.py`) with a clear error: `pip install \"trajectory-ir[dbos]\"`.
3. You already added `dbos = [\"dbos\"]` and put dbos under `dev`  good  but also update any docs/QUICKSTART that assume DBOS is always installed, and make sure e2e/conformance extras still pull `dbos` (via `dev` or an explicit `.[dbos]` in CI).

### Verdict
**Request changes**  finish the lazy-import + install-path story, then this is the right fix for #206.

PR #245, opened 2026-08-17
Fix: make rfc8785 import optional

### LINKED ISSUE #209: bug: PyTest collection fails when `rfc8785` is missing, even in dry-runs

desireddymohithreddy0925:
Resolves #209 by catching ImportError when loading rfc8785. A RuntimeError is now raised lazily only if hashing/sizing functions are called without the dependency installed, ensuring PyTest collection does not crash when isolating driver tests.
Fixes #209.

This PR makes the `rfc8785` import optional at the module level in `nodes.py` and `projector.py`. If the library is missing, a `RuntimeError` is raised lazily only when actual payload hashing or projection sizing is invoked.

This ensures that `pytest` collection does not crash immediately upon scanning the modules in environments where only specific driver tests are isolated and core dependencies might not be fully installed.

### DIFF
--- a/go/trajir/durable/sqlite.go
+++ b/go/trajir/durable/sqlite.go
@@ -27,6 +27,11 @@ func OpenLocal(path string) (*LocalSQLite, error) {
 	}
 	db.SetMaxOpenConns(1)
 
+	if _, err := db.Exec(`PRAGMA journal_mode=WAL;`); err != nil {
+		_ = db.Close()
+		return nil, fmt.Errorf("durable sqlite wal: %w", err)
+	}
+
 	schema := `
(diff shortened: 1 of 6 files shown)


Ayush-Patel-56:
CI failure, can u please fix it

jakharmonika364:
CI is still failing

desireddymohithreddy0925:
@jakharmonika364 could you please give some time, I'll fix all the errors

Ayush-Patel-56:
CI is still failing, please fix it

PR #247, opened 2026-08-17
Fix: catch ImportError for optional DBOS backend

### LINKED ISSUE #232: bug: Moving DBOS to optional dependencies unconditionally breaks `trajectory_client`
### What happened

Since `dbos` was moved to `[project.optional-dependencies]`, importing `client.python.trajectory_client` now unconditionally crashes with a `ModuleNotFoundError`. The client globally and unconditionally imports `drivers.durable_backend.dbos.adapter` on initialization, rendering the base package broken for users who do not install the `[dbos]` extra.

### What you expected

The `trajectory_client` should defer importing backend-specific adapters until they are explicitly initialized by the user, or it should safely catch `ImportError`.

### Steps to reproduce

1. `pip install -e .` (do NOT include `[dbos]`)
2. `python -c "import client.python.trajectory_client"`
3. Observe the crash.

### Environment

Any OS, Python 3.11+

### Logs or output

```text
ModuleNotFoundError: No module named 'dbos'
```

client/python/trajectory_client.py

desireddymohithreddy0925:
Resolves #232 by gracefully catching ImportError when importing the DBOS adapter in the client SDK. It falls back to a no-op init_backend if the [dbos] extra is not installed, preventing unconditional crashes.
Fixes #232.

This PR gracefully catches `ImportError` when importing the DBOS adapter at the root level of `client/python/trajectory_client.py`. Since DBOS was moved to optional dependencies, unconditional imports caused the client SDK to crash for users without the `[dbos]` extra. 

If DBOS is missing, `init_backend` now defaults to a safe no-op, allowing the client SDK to function normally for users who do not rely on the durable execution backend.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
         run: ruff format --check pkg drivers client test conformance examples
 
       - name: Mypy
-        run: mypy pkg/trajectory_ir
+        run: mypy
 
       - name: Cross language hash goldens (Python)
         run: pytest test/unit/test_hash_vectors.py test/unit/test_node_identity.py -q
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -202,7 +202,7 @@ Useful commands:
 ```bash
 ruff check pkg drivers client test conformance examples
 ruff format pkg drivers client test conformance examples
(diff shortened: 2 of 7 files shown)


Ayush-Patel-56:
CI is failing, please fix it

PR #248, opened 2026-08-17
Fix: resolve unclosed file descriptor leak in test harness

### LINKED ISSUE #233: bug: Unclosed file descriptor leak in `_util.py` test harness
### What happened

In `test/e2e/_util.py`, `log_path` is opened with `open(..., "w")` and passed directly to `subprocess.Popen` as `stdout`. Because the file descriptor is never explicitly closed via `log.close()` or wrapped in a context manager, the resource leaks into the OS if the harness crashes or exceptions are raised during execution.

### What you expected

The log file should be opened using a `with open(...) as log:` context manager that encompasses the `proc.wait()` lifecycle.

### Steps to reproduce

1. Open `test/e2e/_util.py`.
2. Locate `log = open(log_path, "w", encoding="utf-8")`.
3. Note the missing `.close()` or `with` block.

### Environment

Any OS, Python 3.11+

### Logs or output

```text
ResourceWarning: unclosed file <_io.TextIOWrapper name='...'>
```

test/e2e/_util.py

desireddymohithreddy0925:
Resolves #233 by turning start_agent into a context manager, allowing the log file to be opened within a 'with open(...)' block that automatically and safely closes the handle when the subprocess lifecycle ends.
Fixes #233.

This PR fixes a file descriptor leak in the `test/e2e/_util.py` test harness where the `log_path` was opened without being properly closed or wrapped in a context manager. 

`start_agent` has been refactored into a context manager using `@contextmanager`. This allows it to safely manage the `open(...)` lifecycle and automatically reap the file descriptor at the end of the block, preventing resource leaks even if the test crashes or an exception is raised mid-execution.

### DIFF
--- a/test/e2e/_util.py
+++ b/test/e2e/_util.py
@@ -2,6 +2,8 @@ import os
 import subprocess
 import sys
 import time
+from collections.abc import Iterator
+from contextlib import contextmanager
 
 REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 AGENT_SCRIPT = os.path.join(REPO_ROOT, "examples", "kill-mid-deploy", "agent.py")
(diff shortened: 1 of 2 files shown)


jakharmonika364:
please fix the failing CI

Issue #175, opened 2026-08-13
process: CNCF Sandbox readiness process track 


Siddhartha-singh01:
## Summary

Tracking issue for the **CNCF Sandbox process** workstream (process and readiness, not product features).

We already have:
- Outline: `docs/CNCF_SANDBOX_APPLICATION_OUTLINE.md`
- Process pack on main: `MAINTAINERS.md`, `GOVERNANCE.md`, `docs/ROADMAP.md`, `ADOPTERS.md`, `docs/SCOPE_AND_NON_GOALS.md`
- Strong CI / protection / releases (Phase 1C + CI/CD harden done)

**We are not ready to apply yet**  mainly repo age (≥ 6 months), Contribution Agreement signatory, trademark agreement, OpenSSF badge, dependency license hygiene, and real adopters.

Related: #172 (priority pack epic), Future #149 (signatures  out of scope here).

## Owner

@Ayush-Patel-56  please read and drive understanding of this process with the team.

### P0  Do not open cncf/sandbox until green

- [ ] Confirm repo age **≥ 6 months** (created ~2026-07-26) before filing
- [ ] Confirm `MAINTAINERS.md` company column is accurate for every maintainer
- [x] Confirm product/company separation wording (`docs/ROADMAP.md` / `docs/SCOPE_AND_NON_GOALS.md`)
- [ ] Name **Contribution Agreement** signatory (legal/org) + application contact emails
- [ ] Maintainers agree trademarks/accounts can transfer to LF if accepted

### P1 Strengthen before TOC review

- [x] Start **OpenSSF Best Practices** badge (achieved: passing -- https://www.bestpractices.dev/projects/14075)
- [x] Dependency license pass vs CNCF third-party allowlist (#254)
- [ ] Fill `ADOPTERS.md` only with real consented entries
- [x] Freeze one-line pitch + 100–300 word description for the form
- [ ] Plan neutral GitHub org name for post-accept transfer

### Apply gate

Only file https://github.com/cncf/sandbox/issues/new?template=application.yml when P0 is complete and pitch is stable.

## References

- https://contribute.cncf.io/projects/submit-project/
- https://github.com/cncf/sandbox/blob/main/README.md
- `docs/CNCF_SANDBOX_APPLICATION_OUTLINE.md`
- Milestone: **CNCF Sandbox prep**

Siddhartha-singh01:
[...]

Issue #167, opened 2026-08-12
ci: SHA-pin GitHub Actions to commit digests


Siddhartha-singh01:
Parent: Phase CI/CD harden epic. Pin uses: lines to immutable SHAs with version comments. Dependabot already updates github-actions. Improves Scorecard Pinned-Dependencies and enables fail-closed zizmor.

PR #254, opened 2026-08-19
chore: add dependency license scan vs CNCF allowlist

### LINKED ISSUE #253: Add dependency license scan vs CNCF allowlist (Python + Go)
## Why

CNCF Sandbox review checks that a project's third-party dependencies use
CNCF-acceptable licenses (permissive: Apache-2.0/MIT/BSD/ISC/PSF freely;
weak-copyleft like LGPL/MPL needs review; GPL/AGPL/SSPL is disallowed).
Tracked as a P1 item on the sandbox-readiness checklist (#175). We have never
actually run this scan against our real dependency tree.

## Scope

Scanned the actual resolved dependency trees (not the ambient dev
environment):

**Python** (`dbos`, `rfc8785`, `psycopg[binary]`, `boto3` + transitives, via a
clean venv install of `pip install .[postgres,s3]`): all permissive
(Apache-2.0 / MIT / BSD / PSF-2.0) **except `psycopg` / `psycopg-binary`,
which are LGPL-3.0-only**. This is an optional extra (`postgres`), imported
as an unmodified library, not statically linked — the standard basis on
which CNCF/ASF-style policy treats LGPL as acceptable — but it should be
recorded explicitly rather than silently passing.

**Go** (`go/go.mod` full resolved tree via `go-licenses`): 100% permissive
(Apache-2.0 / MIT / BSD-3-Clause / ISC). No GPL/AGPL/LGPL anywhere in the
tree. Two modules the tool flagged `Unknown` are tool false-negatives, not
real gaps: `modernc.org/mathutil` has a BSD-3-Clause LICENSE file the
tool's regex missed, and `nexus-rpc/nexus-proto-annotations`'s v0.1.0 tag
predates that repo adding its (confirmed MIT, via GitHub API) LICENSE file.

## What this issue asks for

1. A CI job (Python `pip-licenses --allow-only=... --ignore-packages psycopg
   psycopg-binary` + Go `go-licenses check --disallowed_types=forbidden`)
   so a future dependency bump that pulls in a GPL/AGPL/SSPL license fails
   the PR instead of going unnoticed.
2. A short doc recording the scan methodology, current results, and the
   documented LGPL exception for `psycopg`, so this isn't re-litigated from
   scratch on every CNCF review.

Follow-up to #175.

Ayush-Patel-56:
## Summary

- Adds two new jobs to `.github/workflows/security-scan.yml`: **Dependency license scan (Python)** (`pip-licenses --allow-only=... --ignore-packages psycopg psycopg-binary`) and **Dependency license scan (Go)** (`go-licenses check ./... --disallowed_types=forbidden`), gating every PR on the real resolved dependency tree staying within CNCF's allowed-license policy.
- Adds `docs/THIRD_PARTY_LICENSES.md` documenting the scan methodology and current results.

## Findings

Scanned the actual resolved dependency trees (clean venv `pip install .[postgres,s3]` for Python; full `go/go.mod` tree for Go), not the ambient dev environment.

- **Python**: all permissive (Apache-2.0/MIT/BSD/PSF-2.0) **except `psycopg`/`psycopg-binary`, which are LGPL-3.0-only.** `postgres` is an optional extra, imported as an unmodified library (never statically linked) -- the standard basis CNCF/ASF-style policy treats LGPL as acceptable. Explicitly documented and excluded via `--ignore-packages` rather than silently passing.
- **Go**: 100% permissive (Apache-2.0/MIT/BSD-3-Clause/ISC). No GPL/AGPL/LGPL anywhere. Two modules `go-licenses` flags `Unknown` are verified tool false-negatives (`modernc.org/mathutil` has a BSD-3-Clause LICENSE the tool's regex misses; `nexus-rpc/nexus-proto-annotations`'s pinned tag predates that repo's LICENSE file, confirmed MIT via `gh api`) -- excluded via `--ignore` rather than left unverified.

Fixes #253. Follow-up to #175 (dependency license pass vs CNCF allowlist).

## Test plan

- [x] `python -m yaml.safe_load` on the modified workflow -- valid YAML
- [x] `actionlint` against the modified workflow -- clean
- [x] `zizmor` against the modified workflow -- same finding count per new job as every existing job in this file (advisory-only, `continue-on-error: true`, pre-existing pattern per the file's own #167 TODO)
- [x] Dry-ran both new CI checks locally against the real dependency trees (`pip-licenses --allow-only=...`, `go-licenses check ./... --disallowed_types=forbidden`) -- both pass with the documented exceptions, and I confirmed the Python check correctly fails without the `psycopg` ignore
- Docs-only change to `README.md`/`docs/CI_HARDENING.md`, no source touched -- existing test suite unaffected

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -79,3 +79,64 @@ jobs:
         # Findings reported for review; fail-closed after SHA-pinning (#167).
         continue-on-error: true
         run: zizmor --format plain .github/workflows
+
+  license-scan-python:
+    name: Dependency license scan (Python)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
+        with:
+          python-version: "3.12"
+
+      # Clean venv install of the real runtime dependency tree (dbos, rfc8785,
+      # psycopg[binary], boto3 + transitives) -- not the ambient dev environment,
+      # which pulls in unrelated packages that were never part of this project.
+      - name: Install runtime dependencies
+        run: pip install --upgrade pip pip-licenses ".[postgres,s3]"
+
+      # CNCF allowlist: permissive licenses only. psycopg/psycopg-binary are
+      # LGPL-3.0-only -- an optional extra, imported as an unmodified library
+      # (never statically linked), which is the standard basis on which
+      # CNCF/ASF-style policy treats LGPL as acceptable. Documented exception,
+      # see docs/THIRD_PARTY_LICENSES.md. Anything else outside this list
+      # (e.g. a future GPL/AGPL/SSPL transitive dependency) fails the build.
+      - name: Check licenses against CNCF allowlist
+        run: |
+          pip-licenses \
+            --allow-only="Apache Software License;Apache-2.0;MIT License;MIT;MIT AND PSF-2.0;BSD License;BSD-3-Clause;BSD-2-Clause;ISC License (ISCL);ISC;Python Software Foundation License;PSF-2.0;Apache Software License; BSD License" \
+            --ignore-packages psycopg psycopg-binary
+
+  license-scan-go:
+    name: Dependency license scan (Go)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Set up Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
+        with:
+          go-version: "1.25.x"
+
+      - name: Install go-licenses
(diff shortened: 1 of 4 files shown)


Siddhartha-singh01 (approved):
## Review (approve with nits)

Strong CNCF-aligned work. Scanning the **resolved** dependency trees (not the ambient env) and documenting LGPL for optional `psycopg` is the right honesty bar for Sandbox prep (#175 / #253). CI is green including the new license jobs. Docs + README link are clear.

### What works
- Fail-closed allowlist for Python via `pip-licenses --allow-only=...`
- Explicit `--ignore-packages` for the documented LGPL exception (not silent)
- Go `go-licenses check` with `--disallowed_types=forbidden` and documented Unknown false-negatives
- Actions already SHA-pinned for checkout/setup-python/setup-go

### Nits before or right after merge
1. **`go install github.com/google/go-licenses@latest`**  Scorecard will flag unpinned go install. Prefer a semver tag (e.g. `@v1.6.0`) like we do for actionlint/govulncheck.
2. **Python install step** still uses unhashed `pip install` consistent with current security-scan.yml, but if we land hashed CI locks later, fold these jobs in.
3. Confirm whether the new job names should become **required** checks on `main` (optional; can wait until they stay green for a week).

### Verdict
**Approve.** Ship it; nits can be a tiny follow-up.

PR #255, opened 2026-08-19
fix: sweep stale temp files on fs CAS init (#237)

### LINKED ISSUE #237: bug: `fs.py` CAS backend fails to handle `SIGKILL` temp file accumulation
### What happened

In `pkg/trajectory_ir/storage/fs.py`, atomic writes use `tempfile.mkstemp` to create `.hash.` prefixed temporary files. While standard exceptions clean up the file, if the process receives a `SIGKILL` or power failure before `os.replace` executes, the temp files are permanently orphaned in the CAS directory because there is no startup garbage collection.

### What you expected

The filesystem driver should perform a quick sweep on initialization to delete stale `.*.` temporary files older than a certain threshold.

### Steps to reproduce

1. Hard kill (SIGKILL) a process during a large CAS write.
2. Inspect the CAS directory.
3. Observe orphaned temporary files accumulating.

### Environment

Any OS, Python 3.11+

### Logs or output

```text
N/A
```

pkg/trajectory_ir/storage/fs.py

ujwal-224:
## Summary

In `pkg/trajectory_ir/storage/fs.py`, atomic writes use `tempfile.mkstemp` to create `.{hash}.` prefixed temporary files. If the process receives a SIGKILL or power failure before `os.replace` executes, the temporary files are permanently orphaned in the CAS directory because there is no startup garbage collection.

This PR adds a `_sweep_stale_temp_files` method that runs on `FileSystemCAS` initialization. It recursively sweeps the root directory for temporary files and deletes any that are older than 24 hours (86400 seconds).

## Related issues

Closes #237

## How I tested

```bash
# Python reference (if you touched Python)
pip install -e ".[dev]"
pytest test/unit -q
pytest test/e2e -q
pytest conformance/ -q
ruff check pkg drivers client test conformance examples

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,10 @@ addopts = [
     "-ra",
     "--strict-markers",
 ]
+markers = [
+    "integration: marks tests as integration tests",
+    "e2e: marks tests as end-to-end tests",
+]
 
 [tool.ruff]
 line-length = 100


Siddhartha-singh01 (changes requested):
Thanks for picking up #237. Startup GC for orphaned CAS temps is a real gap and the intent matches how put() writes (mkstemp with prefix `.{hash}.` then os.replace).

Not ready to merge yet.

## Must fix

1. **No tests**
   `test/unit/test_fs_cas.py` already covers FileSystemCAS and this PR adds behavior with zero assertions. Please add at least:
   - plant a fake stale temp under `cas/<shard>/` with mtime older than 24h, construct FileSystemCAS, assert it is gone
   - plant a fresh temp (mtime now), construct FileSystemCAS, assert it is kept
   - plant a real object (put or write a final hash file), assert sweep never deletes it

2. **Sweep scope is too wide**
   ```python
   for p in self._root.rglob(".*.*"):
   ```
   Temps from put() only live under `self._root / "cas" / <2hex> /`. Sweeping the entire root means any other multi-dot hidden file under root (for example `.env.local` if someone pointed root at a broader directory) can be deleted after 24h. Scope the walk to the `cas/` tree, e.g. `(self._root / "cas").rglob(...)` and no-op if `cas/` does not exist yet.

3. **Matcher is looser than the writer**
   Writer prefix is `.{64-hex}.`. Prefer matching that shape (name starts with `.`, then 64 hex, then `.`) instead of generic `.*.*`. Same bug class as over-deleting, just tighter.

## Should fix / follow up

4. **Go parity**
   `go/trajir/cas/cas.go` uses the same pattern (`CreateTemp(..., "."+h+".*")`) and still has no startup sweep. Phase 1B is Go primary. Either add the same GC in this PR or open a linked follow up issue and mention it in the PR body. Do not leave Python fixed and Go forever orphaning temps without a tracker.

5. **PR body / commits**
   Testing section and checklist look truncated. Second commit is a merge from main with an empty body. Fine if DCO is green, but please finish the checklist boxes you claim.

## What looks fine

- Age gate (86400s) is a reasonable default so in-flight puts are not deleted.
- Best effort `OSError` swallow is OK for GC.
- Calling sweep from `__init__` after mkdir matches the issue ask.
- CI is green on the current patch (expected: no failing tests were added).

Once (1) and (2) are in (and ideally (3)), happy to re review.

PR #256, opened 2026-08-19
fix(log): scope has/Has by tenant_id to enforce multi-tenant isolation

### LINKED ISSUE #226: bug: NodeLog.has() and PostgresNodeLog.has() query without tenant_id, violating multi-tenant isolation
## What happened

`NodeLog.has()` (Python SQLite/Postgres) and `NodeLog.Has()` (Go SQLite/Postgres) perform queries without filtering by `tenant_id`:

```sql
SELECT 1 FROM nodes WHERE trajectory_id = ? AND step_n = ? AND kind = ? [AND seq = ?]
```

Across the codebase, the schema (`idx_nodes_slot` on `(tenant_id, trajectory_id, step_n, seq, kind)`), `Append`, `ClaimToolCall`, and `ListNodes` enforce `tenant_id` scoping to guarantee multi-tenant isolation. However, `has` / `Has` is the only read method that omits `tenant_id`.

In multi-tenant environments where multiple tenants share a database instance (or a shared SQLite/PostgreSQL log), if Tenant A and Tenant B use overlapping trajectory IDs (e.g. `t1` or `run-1`), a `has()` query executed for Tenant B returns `True`/`true` if Tenant A has recorded that step, leaking execution presence across tenant boundaries.

### Affected locations

- `pkg/trajectory_ir/runtime/log.py` (`has`, line 180)
- `drivers/postgres/log.py` (`has`, line 264)
- `go/trajir/log/log.go` (`Has`, line 213)
- `go/trajir/postgres/log.go` (`Has`, line 260)

## What you expected

`has` / `Has` should accept `tenant_id` (or filter by `tenant_id`) to maintain strict tenant isolation across all storage read/write operations, matching `list_nodes`, `append`, and `claim_tool_call`.

## Steps to reproduce

In Python:
```python
from trajectory_ir.runtime.log import NodeLog

log = NodeLog(":memory:")
# Tenant A appends a node
log.append("DECISION", 1, {"plan": "A"}, "t1", "tenant_a", 1)

# Tenant B queries if trajectory "t1" has DECISION
# Expected: False (tenant_b never wrote to t1)
# Actual: True (leaks tenant_a's record)
print(log.has("t1", 1, "DECISION"))
```

In Go:
```go
nl, _ := nodelog.Open(":memory:")
step := 1
nl.Append("DECISION", &step, map[string]any{"plan": "a"}, "t1", "tenant-a", 1)

// Tenant B query leaks Tenant A's presence:
ok, _ := nl.Has("t1", 1, "DECISION", nil) // returns true!
```

## Environment

Windows 11 / Linux, Python 3.11+, Go 1.25+

## Logs or output

```text
Tenant B query result: True
```

## Related spec section

README §4 (Content Addressed Append-Only Log), infrastructure.md §1.2 (Multi-tenant schema definition), SECURITY.md (`tenant_id` filter requirements).

Krirox:
## Summary

NodeLog.has() and PostgresNodeLog.has() queried by trajectory_id, step_n, and kind without filtering on tenant_id. In a shared database with overlapping trajectory IDs across tenants, one tenant's nodes could satisfy another tenant's existence check, leaking execution state across tenant boundaries.

This PR makes tenant_id a required positional parameter in has() / Has() across all four NodeLog implementations (Python SQLite, Python Postgres, Go SQLite, Go Postgres), adds has_all_tenants() / HasAllTenants() as an explicit escape hatch for diagnostics, validates that tenant_id is non-empty, and fixes the Go Postgres driver's hard-coded placeholder for seq to use dynamic numbering.

Note: This is a breaking signature change for has() / Has() (now accepts tenant_id as second positional parameter).

## Related issues

Closes #226

## How I tested

`ash
# Python reference
pip install -e .[dev]
pytest test/unit -q
pytest conformance/ -q
ruff check pkg drivers client test conformance examples
mypy

# Go primary
go vet ./...
`

All 169 Python tests pass (11 skipped for live services). All Go tests pass across all packages.

## Checklist

- [x] Commits are DCO signed (\git commit -s\)
- [ ] If you changed \.github/workflows/**\, note it under Safety and expect extra review

## Safety areas

- [x] No

## AI assistance (if any)

None.

### DIFF
--- a/conformance/r03_pure_recompute_test.py
+++ b/conformance/r03_pure_recompute_test.py
@@ -73,7 +73,7 @@ def test_r03_pure_tool_may_recompute_after_prior_tool_call(tmp_path) -> None:
         assert ei.value.step_n == 1
         assert ei.value.tool_name == "compute"
         assert calls["n"] == 2
-        assert log.has(traj, 1, "ABORT", seq=3)
+        assert log.has(traj, tenant, 1, "ABORT", seq=3)
     finally:
         log.close()
 
@@ -105,7 +105,7 @@ def test_r03_pure_run_step_does_not_block_on_second_invocation(tmp_path, monkeyp
     assert r1 == [1]
     assert r2 == [2]
     assert _R03_CALLS["n"] == 2
-    assert log.has(traj, 1, "TOOL_CALL", seq=2)
-    assert log.has(traj, 2, "TOOL_CALL", seq=2)
-    assert not log.has(traj, 1, "ABORT", seq=3)
-    assert not log.has(traj, 2, "ABORT", seq=3)
+    assert log.has(traj, tenant, 1, "TOOL_CALL", seq=2)
+    assert log.has(traj, tenant, 2, "TOOL_CALL", seq=2)
+    assert not log.has(traj, tenant, 1, "ABORT", seq=3)
+    assert not log.has(traj, tenant, 2, "ABORT", seq=3)
--- a/docs/E2E_POSTGRES_CAS_THIN.md
+++ b/docs/E2E_POSTGRES_CAS_THIN.md
@@ -235,7 +235,7 @@ from drivers.postgres.log import open_postgres_node_log
 log = open_postgres_node_log()  # reads TRAJIR_DATABASE_URL
 traj, tenant = "e2e-pg-1", "demo"
 log.append("DECISION", 1, {"plan": {"tool_calls": []}}, traj, tenant, seq=1)
-assert log.has(traj, 1, "DECISION")
+assert log.has(traj, tenant, 1, "DECISION")
 rows = log.list_nodes(traj, tenant_id=tenant)
 print("count", len(rows), "id", rows[0]["id"][:16])
 log.close()
--- a/drivers/postgres/log.py
+++ b/drivers/postgres/log.py
@@ -260,9 +260,50 @@ class PostgresNodeLog:
                 self._conn.rollback()
                 raise
 
-    def has(self, trajectory_id: str, step_n: int, kind: str, seq: int | None = None) -> bool:
+    def has(
+        self,
+        trajectory_id: str,
+        tenant_id: str,
+        step_n: int,
+        kind: str,
+        seq: int | None = None,
+    ) -> bool:
+        """Does a node of `kind` exist for this trajectory/step scoped to `tenant_id`?
+
+        Matches SQLite NodeLog semantics. Callers needing cross-tenant existence
+        checks must use `has_all_tenants` explicitly.
+        """
+        if not tenant_id:
+            raise ValueError("tenant_id must be a non-empty string")
+        return self._has(trajectory_id, tenant_id=tenant_id, step_n=step_n, kind=kind, seq=seq)
+
+    def has_all_tenants(
+        self,
+        trajectory_id: str,
+        step_n: int,
+        kind: str,
+        seq: int | None = None,
+    ) -> bool:
+        """Does a node of `kind` exist for this trajectory/step across any tenant?
+
+        Bypasses tenant isolation for administrative diagnostics and debugging.
+        """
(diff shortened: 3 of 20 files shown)


Siddhartha-singh01 (changes requested):
Thanks for taking #226 end to end. The direction is right and it matches how we already treat ListNodes / list_nodes: default path is tenant scoped, cross tenant needs an explicit AllTenants / all_tenants escape hatch. Dual SDK parity, call site updates, the Go Postgres dynamic $N fix for seq, and CI looking green are all good.

Requesting a few changes before merge.

## Must fix

1. **Tenant leak regression is not actually covered**
   Issue #226 is: tenant A writes DECISION for traj `t1`, tenant B asks `has(t1, …)` with no seq and gets true.
   The new `test_has_tenant_filter` / `TestHasTenantFilter` appends rows for both `tenant-a` and `tenant-b`, then checks seq and a third tenant. That never asserts the real leak.

   Please add (Python SQLite + Postgres unit, and Go SQLite at minimum):

   ```text
   append DECISION for t1 / tenant-a only
   assert has(t1, tenant-b, 1, DECISION) is false
   assert has(t1, tenant-a, 1, DECISION) is true
   ```

   Without that, we can regress the bug again and still pass the suite.

2. **`has_all_tenants` docstring is wrong**
   In `pkg/trajectory_ir/runtime/log.py` it says callers should "perform their own tenant-scoping check on the result".
   `has` / `has_all_tenants` returns a bool. There is no result set to filter. That text was clearly copied from `list_nodes_all_tenants`. Same smell on the Go `HasAllTenants` comment if it implies the same. Please rewrite for a boolean existence check (and say when it is legitimate to use, e.g. admin/debug only, not normal host paths).

## Should fix

3. **No test for the escape hatch**
   You added `has_all_tenants` / `HasAllTenants` but never call it in tests. One small case is enough: two tenants write the same traj/step/kind, scoped `has` for a third tenant is false, `has_all_tenants` is true.

4. **Empty `tenant_id`**
   `Has(traj, "", …)` / `has(traj, "", …)` still runs a scoped query against the empty string instead of failing closed. Prefer rejecting empty tenant on the scoped API (same idea as requiring tenant on ListNodes callers). If you intentionally want to match ListNodes behavior exactly, say so in the PR and we can accept parity, but fail closed is better for a security fix.

## Nits (non blocking)

5. Postgres Python `has` has no docstring while the SQLite one does. One line parity would help.
6. Call this out as a breaking signature change in the PR body (fine for where we are, just so reviewers and early adopters see it).

## What looks solid

- Aligns with README tenant_id in the node identity / schema story. This is isolation in the log API, not building a multi tenant SaaS product.
- Shared `_has` / `has` helpers are clean.
- Go Postgres `seq = $4` bugfix when tenant is injected is a real footgun catch.
- Conformance, client, resume, docs snippet, and fakes were updated consistently.
- Checklist / DCO / local test claims match what CI is showing.

Once (1) and (2) are in, I am happy to re review quickly.

Siddhartha-singh01 (approved):
Re reviewed after the follow up push. The earlier blockers are addressed.

What landed from the last review:
1. Real #226 leak regression: tenant-a writes, tenant-b has(...) is false (Python SQLite, Postgres unit, Go SQLite).
2. has_all_tenants / HasAllTenants docs now describe a bool existence check for admin/diagnostic use, not a fake result set filter.
3. Escape hatch covered by test_has_all_tenants / TestHasAllTenants.
4. Empty tenant_id rejected on the scoped API (ValueError / error).
5. Breaking signature called out in the PR body.
6. Postgres Python has docstring parity.

CI is green. Happy to approve.

Non blocking nits only:
- Whitespace only tenant_id (e.g. " ") still passes the empty check in both languages. Trimming would be slightly stricter if you touch this again later.
- Go Postgres package does not duplicate TestHasAllTenants / TestHasRejectsEmptyTenant (SQLite does). Logic is the same one liner, so fine for now.

LGTM for merge from my side.

jakharmonika364 (approved):
LGTM, Thanks!

PR #259, opened 2026-08-20
fix(cas): sweep stale put temps on filesystem CAS init

### LINKED ISSUE #237: bug: `fs.py` CAS backend fails to handle `SIGKILL` temp file accumulation

Siddhartha-singh01:
## Summary

Closes #237.

put() / Put write with a `.{hash}.` temp then rename. Hard kill can leave those temps behind. This adds an **explicit** stale temp sweep (not on by default in the constructor).

## What it does

- `sweep_stale_temp_files()` / `SweepStaleTempFiles(maxAge)` walk only `cas/`
- Deletes names matching `.{lowercase-64-hex}.…` older than 24h (default)
- Python opt-in: `FileSystemCAS(root, sweep_stale_temps=True)`
- Default construct does **not** sweep (safe for shared roots / concurrent Put)
- Dual SDK (Python + Go)

## Test plan

- [x] `PYTHONPATH=pkg pytest test/unit/test_fs_cas.py -q`
- [x] `go test ./trajir/cas/ -count=1`
- [ ] CI green on latest commit

## Review follow up

Addressed Monica (no ctor race, walk OSError) and Ayush (lowercase hex matches ContentHash).

### DIFF
--- a/go/trajir/cas/cas.go
+++ b/go/trajir/cas/cas.go
@@ -18,6 +18,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 )
 
 // Sentinel errors.
@@ -29,6 +30,11 @@ var (
 )
 
 var hex64 = regexp.MustCompile(`^[0-9a-f]{64}$`)
+
+// Matches CreateTemp names: ".{64-hex}." + random suffix (same as Python mkstemp).
+var tempNameRe = regexp.MustCompile(`^\.[0-9a-f]{64}\.`)
+
+const defaultTempMaxAge = 24 * time.Hour
 
 // Store is the minimal CAS surface used by thin package rehydrate.
 type Store interface {
@@ -78,7 +84,36 @@ func NewFileSystem(root string) (*FileSystem, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &FileSystem{Root: abs}, nil
+	fs := &FileSystem{Root: abs}
+	fs.sweepStaleTempFiles(defaultTempMaxAge)
+	return fs, nil
+}
+
+// sweepStaleTempFiles removes orphaned Put temp files under cas/ older than maxAge.
+// Only names matching the CreateTemp prefix are considered.
+func (fs *FileSystem) sweepStaleTempFiles(maxAge time.Duration) {
+	casRoot := filepath.Join(fs.Root, "cas")
+	st, err := os.Stat(casRoot)
+	if err != nil || !st.IsDir() {
+		return
+	}
+	now := time.Now()
+	_ = filepath.WalkDir(casRoot, func(path string, d os.DirEntry, err error) error {
+		if err != nil || d.IsDir() {
+			return nil
+		}
+		if !tempNameRe.MatchString(d.Name()) {
+			return nil
+		}
+		info, err := d.Info()
+		if err != nil {
+			return nil
+		}
+		if now.Sub(info.ModTime()) > maxAge {
+			_ = os.Remove(path)
+		}
+		return nil
+	})
 }
 
 // PathFor returns the absolute path for a content hash under this store.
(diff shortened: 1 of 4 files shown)


Siddhartha-singh01:
Ready for a look. Scoped the sweep to cas/ only, matched the real temp name shape, dual SDK, and added tests. Closes #237.

go/trajir/cas/cas.go
```diff
@@ -30,6 +31,11 @@ var (
 
 var hex64 = regexp.MustCompile(`^[0-9a-f]{64}$`)
 
+// Matches CreateTemp names: ".{64-hex}." + random suffix (same as Python mkstemp).
+var tempNameRe = regexp.MustCompile(`^\.[0-9a-f]{64}\.`)
```
Ayush-Patel-56, line 37: why we are using only lowerCase here not mix of both upper & lower case ? if it's for the purpose, then what's that ?

go/trajir/cas/cas.go
```diff
@@ -78,7 +84,36 @@ func NewFileSystem(root string) (*FileSystem, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &FileSystem{Root: abs}, nil
```
jakharmonika364, line 81: Constructing the store now does an unconditional full-tree walk + delete, with no lock against a concurrent `Put()`. We already reconstruct `FileSystemCAS`/`FileSystem` against shared roots elsewhere in this repo (e.g. `test_adoption_host_example.py:129`), so this isn't hypothetical, a slow write whose temp file crosses 24h can get deleted mid-write by another process's constructor call. Can we gate this behind an explicit `Sweep()`/flag instead of running it unconditionally in the constructor, and/or add a lock or liveness check before deleting?

pkg/trajectory_ir/storage/fs.py
```diff
@@ -39,6 +45,31 @@ class FileSystemCAS:
...
+        put() writes with prefix ``.{hash}.`` then os.replace. A hard kill can leave
+        those temps behind. Only the cas/ tree is scanned, and only names that match
+        the writer prefix, so unrelated hidden files under root are left alone.
+        """
+        cas_root = self._root / "cas"
+        if not cas_root.is_dir():
+            return
+        now = time.time()
+        for path in cas_root.rglob("*"):
+            if not path.is_file():
```
jakharmonika364: The `try/except OSError` only wraps `stat()`/`unlink()`, not the `rglob("*")` iteration or `is_file()`. If a concurrent process deletes something mid-walk, this raises uncaught and crashes `__init__`. Go's `WalkDir` swallows this per-entry, Python needs the same.

jakharmonika364 (changes requested):

Siddhartha-singh01:
Pushed a follow up for the review notes.

@jakharmonika364
- Default construct no longer walks or deletes. Call `sweep_stale_temp_files()` / `SweepStaleTempFiles`, or pass `sweep_stale_temps=True` in Python when you know nothing else is writing that root.
- Python walk now catches OSError around the rglob listing and each entry (same idea as Go WalkDir skipping bad entries).

@Ayush-Patel-56
- Lowercase only on purpose: `content_hash` / `ContentHash` always emit lowercase hex, and put/CreateTemp use that string in the temp prefix. Matching A-F would not hit real writer temps. Left a short comment in both places.

Siddhartha-singh01:
Python Quality was failing on ruff format for the FileSystemCAS __init__ line. Reformatted and pushed.

jakharmonika364 (approved):
Looks solid

Issue #258, opened 2026-08-19
bug: non-atomic file replacement in rewriteZipWithSignature causes package deletion on rename failure


Krirox:
### What happened

In `go/trajir/tir/signature.go`, `rewriteZipWithSignature` replaces the original archive with the signed archive. If the initial `os.Rename(tmpName, path)` fails on Windows (which does not allow overwriting an existing file directly), the fallback code calls `os.Remove(path)` before attempting `os.Rename(tmpName, path)` a second time:

```go
if err := os.Rename(tmpName, path); err != nil {
	// Windows: rename over existing may fail — remove target then rename.
	if rmErr := os.Remove(path); rmErr != nil {
		return fmt.Errorf("%w: replace package: %v (also: %v)", ErrSignature, err, rmErr)
	}
	if err := os.Rename(tmpName, path); err != nil {
		return fmt.Errorf("%w: replace package: %v", ErrSignature, err)
	}
}
```

If the second `os.Rename` fails (due to a transient lock, antivirus scanner, permissions, or cross-device handle issue), the original `.tir` file has already been deleted by `os.Remove(path)`. The function returns an error, and the original file is permanently lost on disk with no recovery path.

### What you expected

Signing a package should be atomic or fail-safe. If replacing the package file fails, the original `.tir` archive should remain intact. A backup rename (e.g. `path + ".bak"`) or transactional rollback should be used so the original archive is restored if the replacement rename fails.

### Steps to reproduce

1. Export a `.tir` archive (`pkg.tir`).
2. Call `tir.Sign("pkg.tir", key, meta)` on Windows.
3. If the second `os.Rename` fails after `os.Remove(path)`, the original `pkg.tir` is destroyed.

### Environment

- OS: Windows 11 / Windows Server
- Go: go1.22+
- Package: `go/trajir/tir`

### Logs or output

```text
trajir-pkg-sig-v1: replace package: rename C:\...\tmp.tir C:\...\pkg.tir: Access is denied.
```
Target file `pkg.tir` is missing from the filesystem.

### Related spec section (optional)

`go/trajir/tir/signature.go` / package signing workflow

Siddhartha-singh01:
Folded into #260 with the backup/restore replace path.

Issue #191, opened 2026-08-14
`Load` and `LoadUnverified` decompress every zip member twice


jakharmonika364:
`go/trajir/tir/tir.go`, around line 714

`loadImpl` already reads every zip member once through `readMember` and `readAny`, bounded by the `128 MiB` budget. But then it unconditionally calls `verifySignatureFromZipFiles`, which calls `collectZipMembers` and re-opens and re-decompresses every entry again into a separate map with its own `128 MiB` budget, even when there's no SIGNATURE member in the package at all.

For packages near the size limit this doubles decompression CPU and roughly doubles peak memory on every single load. Would be good to short circuit early when there's no SIGNATURE member present, or share the already decompressed data between the two passes instead of reading twice.

desireddymohithreddy0925:
This looks like a worthwhile optimization in the loading path. I’d address it by avoiding the second ZIP traversal when the package does not contain a `SIGNATURE` member, while preserving the existing verification behavior for signed packages.

The change should eliminate redundant decompression and reduce unnecessary memory usage for packages approaching the `128 MiB` limit. I’ll also exercise the relevant load and signature-verification paths to ensure the optimization does not alter existing behavior.

desireddymohithreddy0925:
@maintainers To proceed with my implementation please assign me

Ayush-Patel-56:
no need of assignment, go ahead and raise the PR. Thanks!

Issue #190, opened 2026-08-14
Signing changes file permissions from 0644 to 0600 without documentation


jakharmonika364:
`go/trajir/tir/signature.go`, around line 382

Export creates the destination file normally, which ends up around `0644` depending on `umask`. But when `SignKey` is set, Sign rewrites the file through `os.CreateTemp`, which always creates files at 0600, then renames over the destination. So a signed export ends up owner only readable while an unsigned export of the same content does not.

This is a behavior change that isn't documented anywhere, and could break setups where a different user or service account needs read access to exported packages. Worth deciding if this is intentional, and if so, documenting it. If not, we should preserve the original file mode across the rewrite.

Ayush-Patel-56:
This was already fixed in cb474a8 (PR #264, "signature P2 preserve mode + skip unsigned double Load"), merged 2026-08-22.

rewriteZipWithSignature now stats the destination file before creating the temp file, preserves its mode, and chmods the temp file back to that mode before the atomic rename - so a signed export keeps the same permissions as an unsigned one:

Fix: [go/trajir/tir/signature.go:384-390,437-439](https://github.com/Coder-s-OG-s/Trajectory-IR/blob/main/go/trajir/tir/signature.go#L384-L439)

Regression test: TestSignPreservesPackageMode in [go/trajir/tir/signature_test.go:328-357](https://github.com/Coder-s-OG-s/Trajectory-IR/blob/main/go/trajir/tir/signature_test.go#L328-L357)
Closing as resolved

Issue #189, opened 2026-08-14
Missing `fsync` before rename in signature rewrite


jakharmonika364:
`go/trajir/tir/signature.go`, around line 422

The temp file gets closed and renamed over the target without a `tmp.Sync()` call first. This is inconsistent with how `go/trajir/cas/cas.go` does the same temp-then-rename pattern in Put, which does sync before close and rename.

Without the sync, a crash or power loss right after the rename can leave the package truncated even though the rename itself reported success. Should be a small fix, just add the sync call in the same place `cas.go` does it.

desireddymohithreddy0925:
I’ll bring the signature rewrite path in line with the existing temp-file handling used in `cas.go` by syncing the temporary file with `tmp.Sync()` before closing and renaming it.

This should ensure the rewritten package contents are flushed to stable storage before the atomic rename, reducing the risk of corruption or truncation if a crash occurs immediately afterward. I’ll also verify the existing signature rewrite tests to make sure the change integrates cleanly.

desireddymohithreddy0925:
@maintainers To proceed with my implementation please assign me

Ayush-Patel-56:
Thanks for the offer to help @desireddymohithreddy0925 - but this one's already been fixed, so no need to pick it up.

tmp.Sync() was added in 0aac337 (PR #263, "signature P1 fsync + README 9.1 honesty"), merged 2026-08-20. The temp file is now synced before Close() and the atomic rename, matching the pattern in cas.go:

Fix: [go/trajir/tir/signature.go:431](https://github.com/Coder-s-OG-s/Trajectory-IR/blob/main/go/trajir/tir/signature.go#L431)
Closing as resolved. Please reopen if you can still reproduce truncation on a current build.

PR #260, opened 2026-08-20
fix(go): signature P0 safe replace and fail-closed Export SignKey

### LINKED ISSUE #198: fix(go): signature safety P0 after Phase B (destroy package + silent skip sign)
## Why this is next (step 2)

Phase B Go Sign/Verify is **merged** (#178 / #177). Several follow-up bugs need fixing **before** Python Phase C copies the same patterns.

## P0 (do first)

| Issue | Problem |
|-------|---------|
| [#186](https://github.com/Coder-s-OG-s/Trajectory-IR/issues/186) | Rename fallback in `rewriteZipWithSignature` can destroy the original `.tir` |
| [#188](https://github.com/Coder-s-OG-s/Trajectory-IR/issues/188) | Export silently skips signing on bad key length instead of erroring |

## Acceptance

- [ ] #186 fixed + tests (no destroy-on-failed-rename)
- [ ] #188 fixed + tests (bad key length fails closed)
- [ ] Security-minded review on rewrite path
- [ ] Close this tracker when both children are done

## Related later (not this issue)

P1/P2: #189 fsync, #187 README vs impl, #190 permissions, #191 double decompress

Siddhartha-singh01:
## Summary

Closes #198 (and the child bugs #186 / #188). Also covers #258 which is the same destroy-on-failed-rename story.

## Fixes

1. **Safe replace** for `rewriteZipWithSignature`
   - Try normal rename first
   - If that cannot overwrite (Windows), move the original aside, put the temp in place, then drop the backup
   - If the final rename fails, restore the backup so the original `.tir` is not lost

2. **Export SignKey**
   - Empty / unset key: unsigned export (unchanged)
   - Non-empty wrong length (e.g. 32 byte seed): error, no silent skip

## Test plan

- [x] `go test ./trajir/tir/ -count=1`
- [x] `TestReplaceFileViaBackupRestoresOriginal`
- [x] `TestExportRejectsWrongLengthSignKey`
- [ ] CI green

## Out of scope (later P1/P2)

fsync (#189), README honesty (#187), file mode docs (#190), double decompress (#191)

### DIFF
--- a/go/trajir/tir/signature.go
+++ b/go/trajir/tir/signature.go
@@ -422,15 +422,42 @@ func rewriteZipWithSignature(path string, members map[string][]byte, signatureJS
 	if err := tmp.Close(); err != nil {
 		return err
 	}
-	if err := os.Rename(tmpName, path); err != nil {
-		// Windows: rename over existing may fail — remove target then rename.
-		if rmErr := os.Remove(path); rmErr != nil {
-			return fmt.Errorf("%w: replace package: %v (also: %v)", ErrSignature, err, rmErr)
-		}
-		if err := os.Rename(tmpName, path); err != nil {
-			return fmt.Errorf("%w: replace package: %v", ErrSignature, err)
-		}
+	if err := replaceFile(tmpName, path); err != nil {
+		return err
 	}
 	success = true
+	return nil
+}
+
+// afterBackupHook is an optional test seam run after the original package is
+// moved aside and before the temp is moved into place. Production leaves it nil.
+var afterBackupHook func(dest string)
+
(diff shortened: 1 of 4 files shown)


Siddhartha-singh01:
P0 from the ordered list. Safe replace + Export fail closed on bad SignKey. Ready when you are.

jakharmonika364 (approved):
LGTM

Siddhartha-singh01:
@Ayush-Patel-56 Review this once

Ayush-Patel-56 (approved):
Cheers!

PR #261, opened 2026-08-20
fix(tir): preserve original archive on rewriteZipWithSignature rename failure

### LINKED ISSUE #258: bug: non-atomic file replacement in rewriteZipWithSignature causes package deletion on rename failure

Krirox:
## Summary

Prevent permanent .tir package loss on Windows when rewriteZipWithSignature encounters a rename failure. The function now creates a temporary backup (.bak) before moving the signed temporary archive over the destination path, restoring the original archive if the replacement rename fails and removing the backup upon success.

## Related issues

Closes #258

## How I tested

\\\ash
# Go primary
\\\

## Checklist

- [x] Commits are DCO signed (\git commit -s\)
- [ ] If you changed \.github/workflows/**\, note it under Safety and expect extra review

## Safety areas

- [x] No
- [ ] Yes (call it out here)

## AI assistance (if any)

None.

### DIFF
--- a/go/trajir/tir/signature.go
+++ b/go/trajir/tir/signature.go
@@ -422,15 +422,42 @@ func rewriteZipWithSignature(path string, members map[string][]byte, signatureJS
 	if err := tmp.Close(); err != nil {
 		return err
 	}
-	if err := os.Rename(tmpName, path); err != nil {
-		// Windows: rename over existing may fail — remove target then rename.
-		if rmErr := os.Remove(path); rmErr != nil {
-			return fmt.Errorf("%w: replace package: %v (also: %v)", ErrSignature, err, rmErr)
-		}
-		if err := os.Rename(tmpName, path); err != nil {
-			return fmt.Errorf("%w: replace package: %v", ErrSignature, err)
-		}
+	if err := replaceFile(tmpName, path); err != nil {
+		return err
 	}
 	success = true
+	return nil
+}
+
+// afterBackupHook is an optional test seam run after the original package is
+// moved aside and before the temp is moved into place. Production leaves it nil.
+var afterBackupHook func(dest string)
+
+// replaceFile installs src at dest without deleting dest until src is in place.
+// On platforms where rename cannot overwrite (Windows), dest is moved to a
+// sibling backup first; if the final rename fails, the backup is restored.
+func replaceFile(src, dest string) error {
+	if err := os.Rename(src, dest); err == nil {
+		return nil
+	}
+	return replaceFileViaBackup(src, dest)
+}
+
+func replaceFileViaBackup(src, dest string) error {
+	bak := dest + ".trajir-replace-bak"
+	_ = os.Remove(bak)
+	if err := os.Rename(dest, bak); err != nil {
+		return fmt.Errorf("%w: replace package (move aside): %v", ErrSignature, err)
+	}
+	if afterBackupHook != nil {
+		afterBackupHook(dest)
+	}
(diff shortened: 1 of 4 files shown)


Siddhartha-singh01 (changes requested):
Thanks for picking up #258. Moving the original aside instead of os.Remove is the right direction. A few things before this can merge.

## Line notes

`bakPath := path + ".bak"`
Fine idea. Prefer a more specific suffix like `.trajir-replace-bak` so a user file that already ends in `.bak` is less likely to collide. Also clear or move any leftover bak before `Rename(path, bakPath)`, otherwise a stale bak from a prior crash can make the move aside fail on Windows.

`if bakErr := os.Rename(path, bakPath); bakErr != nil`
Good. Returning here with the original still at `path` is correct.

`_ = os.Rename(bakPath, path)` on the failed final rename
This is the weak spot. If restore fails you drop the error on the floor and the caller only sees the replace error. Check the restore error and include it in the returned message (same idea as the old "also: %v" style).

`_ = os.Remove(bakPath)` after success
OK for the happy path. Ignoring remove errors here is fine.

## Bigger gaps

1. Checklist says tests are included, but the diff is only `signature.go` (+5/-3). Need at least one test that forces the backup path and shows the original bytes come back when the second rename fails.

2. We already have #260 open for this same destroy on rename bug (plus the Export SignKey fail closed piece from #198). #260 extracts a small `replaceFile` helper and adds rollback tests. Please either:
   - close this as covered by #260, or
   - rebase onto #260 / fold into that PR so we do not land two slightly different replace strategies.

## Verdict

Request changes. Intent is good, restore error handling and a real regression test are still missing, and we should not double fix #258 next to #260.

PR #263, opened 2026-08-20
fix(go): signature P1 fsync + README 9.1 honesty

Siddhartha-singh01:
## Summary

Closes #199 (children #189 and #187).

Stacked on #260 (signature P0). Do not merge this before #260.

## Changes

1. **fsync** in `rewriteZipWithSignature` via `tmp.Sync()` before close/replace (matches `cas.Put`).
2. **README** 5 and 9.1: Go Sign/Verify + Export SignKey are shipped; Python parity and R09–R11 stay Future (#179 / #149). Unsigned remains the default.

## Test plan

- [x] `go test ./trajir/tir/ -count=1`
- [ ] CI green

## Note

No merge until #260 is in, or retarget this PR to main after #260 lands.

PR #264, opened 2026-08-20
fix(go): signature P2 preserve mode + skip unsigned double Load

### LINKED ISSUE #200: fix(go): signature polish (file mode docs + single-pass Load) #190 #191
## Why this is next (step 4)

Lower severity signature follow-ups after P0/P1.

## Scope

| Issue | Problem |
|-------|---------|
| [#190](https://github.com/Coder-s-OG-s/Trajectory-IR/issues/190) | Signing changes file permissions 0644 to 0600 without documentation |
| [#191](https://github.com/Coder-s-OG-s/Trajectory-IR/issues/191) | `Load` / `LoadUnverified` decompress every zip member twice |

## Acceptance

- [ ] #190: document intentional mode or preserve mode
- [ ] #191: single-pass load (or justified residual cost)
- [ ] Close this tracker when both are done

Siddhartha-singh01:
## Summary

Closes #200 (children #190 and #191).

P0 (#260) and P1 (#263) are already on main. This branch is rebased onto main.

## Changes

1. **#190** Sign rewrite preserves the original package permission bits (chmod temp before replace). Documented in README 9.1.
2. **#191** `Load` / `LoadUnverified` skip the signature member collect pass when there is no `SIGNATURE` entry (unsigned default path).

## Test plan

- [x] `go test ./trajir/tir/ -count=1`
- [x] `TestSignPreservesPackageMode` (unix)
- [ ] CI green

## Note

When SIGNATURE is present, verify still collects members once via the existing helper (needed for optional members). The expensive double read on the common unsigned path is gone.

### DIFF
--- a/go/trajir/tir/signature.go
+++ b/go/trajir/tir/signature.go
@@ -381,6 +381,12 @@ func collectZipMembers(files []*zip.File) (members map[string][]byte, signature 
 
 func rewriteZipWithSignature(path string, members map[string][]byte, signatureJSON []byte) error {
 	dir := filepath.Dir(path)
+	// Preserve destination mode across CreateTemp (which uses 0600) so signed
+	// exports keep the same permissions as the unsigned zip Export wrote.
+	origMode := os.FileMode(0o644)
+	if st, err := os.Stat(path); err == nil {
+		origMode = st.Mode().Perm()
+	}
 	tmp, err := os.CreateTemp(dir, "tir-sign-*.tmp")
 	if err != nil {
 		return err
@@ -426,6 +432,9 @@ func rewriteZipWithSignature(path string, members map[string][]byte, signatureJS
 	}
 	if err := tmp.Close(); err != nil {
 		return err
+	}
+	if err := os.Chmod(tmpName, origMode); err != nil {
+		return fmt.Errorf("%w: preserve package mode: %v", ErrSignature, err)
 	}
 	if err := replaceFile(tmpName, path); err != nil {
 		return err
--- a/go/trajir/tir/signature.go
+++ b/go/trajir/tir/signature.go
@@ -383,10 +383,11 @@ func rewriteZipWithSignature(path string, members map[string][]byte, signatureJS
 	dir := filepath.Dir(path)
 	// Preserve destination mode across CreateTemp (which uses 0600) so signed
 	// exports keep the same permissions as the unsigned zip Export wrote.
-	origMode := os.FileMode(0o644)
-	if st, err := os.Stat(path); err == nil {
-		origMode = st.Mode().Perm()
+	st, err := os.Stat(path)
+	if err != nil {
+		return fmt.Errorf("%w: stat package for mode: %v", ErrSignature, err)
 	}
+	origMode := st.Mode().Perm()
 	tmp, err := os.CreateTemp(dir, "tir-sign-*.tmp")
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


Ayush-Patel-56:
there are merge conflict, please solve it

Siddhartha-singh01:
Rebased onto main now that #260 and #263 are in. Conflicts cleared, tests still pass locally.

Siddhartha-singh01:
Hey @Ayush-Patel-56 conflicts are already cleared. Branch is rebased on current main, CI is green, and GitHub shows mergeable. Please take another look when you can.

go/trajir/tir/tir.go
```diff
@@ -716,9 +716,13 @@ func loadImpl(path string, verify bool) (*Package, error) {
...
-	sigInfo, err := verifySignatureFromZipFiles(zr.File, VerifyOptions{})
-	if err != nil {
-		return nil, err
+	// Unsigned packages skip the second decompress pass (no SIGNATURE member).
+	var sigInfo *SignatureInfo
+	if _, hasSig := byName[SignatureMemberName]; hasSig {
+		sigInfo, err = verifySignatureFromZipFiles(zr.File, VerifyOptions{})
+		if err != nil {
+			return nil, err
+		}
```
jakharmonika364, line 734: skipping `verifySignatureFromZipFiles` here also skips `collectZipMembers`'s duplicate-member check (i think CWE-436 guard), since that check lives inside the function you're now gating on `hasSig`. Unsigned packages are the default case per the README, so this reopens duplicate-zip-member smuggling for the common path. Verified: a `.tir` with two `manifest.json` entries and no `SIGNATURE` is rejected on `main` but accepted here via `LoadUnverified`.
 I would rather suggest hoisting the duplicate-member check out of `verifySignatureFromZipFiles` so it always runs, independent of whether a signature is present.

go/trajir/tir/signature.go
```diff
@@ -381,6 +381,12 @@ func collectZipMembers(files []*zip.File) (members map[string][]byte, signature
 
 func rewriteZipWithSignature(path string, members map[string][]byte, signatureJSON []byte) error {
 	dir := filepath.Dir(path)
+	// Preserve destination mode across CreateTemp (which uses 0600) so signed
+	// exports keep the same permissions as the unsigned zip Export wrote.
+	origMode := os.FileMode(0o644)
+	if st, err := os.Stat(path); err == nil {
+		origMode = st.Mode().Perm()
+	}
```
jakharmonika364: `origMode` falls back to `0644` if `os.Stat` errors, instead of propagating the error, That's inconsistent with the fail-closed `os.Chmod` error handling a few lines below and could quietly loosen permissions on the signed output.
Minor: narrow triggers, but worth `return err` here for consistency

jakharmonika364 (changes requested):

Siddhartha-singh01:
@jakharmonika364 fixed both notes:

1. `loadImpl` now rejects duplicate zip member names when building `byName`, so unsigned Load/LoadUnverified still get the CWE-436 guard even though we skip the second decompress for packages without SIGNATURE. Added a regression test with a duplicate `manifest.json`.
2. `rewriteZipWithSignature` fails closed if `Stat` of the destination fails, instead of falling back to 0644.

jakharmonika364 (approved):
Both fixed, duplicate-member check now runs unconditionally in `loadImpl`, and `origMode` fails closed on stat error. Tests cover both cases. LGTM.

PR #265, opened 2026-08-20
fix(resume): log TOOL_CALL prior to non-gated tool execution (#262)

### LINKED ISSUE #262: bug: non-gated tool execution failures prevent TOOL_CALL audit logging
### What happened

In `go/trajir/resume/step.go` (`RunStep`) and `go/trajir/client/client.go` (`ExecTool`), non-gated tools (`PURE`, `READ_ONLY`, `IDEMPOTENT_WRITE`) invoke `tool.Fn(args)` before appending nodes to `NodeLog`. If `tool.Fn` returns an error, the function returns immediately without recording the `TOOL_CALL` node in the log:

```go
result, err := tool.Fn(args)
if err != nil {
    return nil, err
}
if _, err := t.log.Append(
    "TOOL_CALL",
    &step,
    map[string]any{"tool": tool.Name, "args": args},
    t.TrajectoryID,
    t.TenantID,
    seq,
); err != nil {
    return nil, err
}
```

The same behavior exists in Python (`pkg/trajectory_ir/resume/step.py` and `client/python/trajectory_client.py`).

When a non-gated tool fails during execution, the resulting `NodeLog` contains `PROJECT_CONTEXT` and `DECISION`, but lacks any record that the tool was attempted.

### What you expected

`TOOL_CALL` should be recorded in the `NodeLog` before executing the tool (or recorded on error alongside failure details), preserving audit history and forensic traceability for failed tool invocations.

### Steps to reproduce

1. Configure a step with a `PURE` or `READ_ONLY` tool whose function returns an error (such as network failure or invalid argument).
2. Execute the step via `RunStep` or `ExecTool`.
3. The call returns an error.
4. Inspect the `NodeLog` via `ListNodes(trajectoryID, tenantID)`.
5. Observe that no `TOOL_CALL` node exists for the failed invocation at sequence slot `seq`.

### Environment

- OS: All platforms (Linux, macOS, Windows)
- Go: go1.22+
- Python: 3.11+
- Packages: `go/trajir/resume`, `go/trajir/client`, `pkg/trajectory_ir/resume`, `client/python`

### Logs or output

```text
NodeLog contains PROJECT_CONTEXT (seq 0) and DECISION (seq 1), but no TOOL_CALL at seq 2.
```

`go/trajir/resume/step.go`, `go/trajir/client/client.go`, README section 8

Krirox:
Non-gated tool executions (PURE, READ_ONLY, IDEMPOTENT_WRITE) in Go RunStep/ExecTool and Python make_run_step/exec_tool previously recorded TOOL_CALL after tool invocation. When a tool returned an error or raised an exception, the function returned before TOOL_CALL was appended, leaving no audit record of the attempted invocation in NodeLog.

Record TOOL_CALL before running the tool, while recording TOOL_RESULT only upon successful completion.

Closes #262

### DIFF
--- a/client/python/trajectory_client.py
+++ b/client/python/trajectory_client.py
@@ -3,7 +3,7 @@ from typing import Any
 
 from drivers.durable_backend.dbos.adapter import init_backend
 from trajectory_ir.effects import requires_block_and_gate
-from trajectory_ir.resume.gate import make_gated_tool_call
+from trajectory_ir.resume.gate import make_gated_tool_call, make_plain_tool_call
 from trajectory_ir.runtime.log import NodeLog
 from trajectory_ir.runtime.sandbox import RunMode, assert_tool_allowed_in_mode, normalize_run_mode
 from trajectory_ir.runtime.tool import Tool
@@ -114,25 +114,16 @@ def exec_tool(trajectory: Trajectory, step_n: int, call: dict, tool: Tool, seq: 
         )
         result = fn(**call["args"])
     else:
-        # Not claim-gated, but still needs IR history audit; gate path
-        # already logs its own TOOL_CALL/TOOL_RESULT nodes.
-        result = tool.fn(**call["args"])
-        log.append(
-            "TOOL_CALL",
-            step_n,
-            {"tool": tool.name, "args": call["args"]},
+        fn = make_plain_tool_call(
+            log,
             trajectory.trajectory_id,
             trajectory.tenant_id,
+            step_n,
             seq=seq,
+            tool_name=tool.name,
+            tool_fn=tool.fn,
         )
-        log.append(
-            "TOOL_RESULT",
-            step_n,
-            {"result": result},
-            trajectory.trajectory_id,
-            trajectory.tenant_id,
-            seq=seq + 1,
-        )
+        result = fn(**call["args"])
     return ToolResult(step_n=step_n, result=result)
 
 
--- a/go/trajir/client/client.go
+++ b/go/trajir/client/client.go
@@ -210,7 +210,6 @@ func (t *Trajectory) ExecTool(stepN, seq int, tool resume.Tool, args map[string]
 	if err := sandbox.AssertToolAllowed(t.Mode, tool.Name, tool.Effect); err != nil {
 		return nil, err
 	}
-	step := stepN
 	if effects.RequiresBlockAndGate(tool.Effect) {
 		fn := resume.MakeGatedToolCall(
 			t.log,
@@ -227,28 +226,17 @@ func (t *Trajectory) ExecTool(stepN, seq int, tool resume.Tool, args map[string]
 		}
 		return &ToolResult{StepN: stepN, Result: result}, nil
 	}
-	result, err := tool.Fn(args)
-	if err != nil {
-		return nil, err
-	}
-	if _, err := t.log.Append(
-		"TOOL_CALL",
-		&step,
-		map[string]any{"tool": tool.Name, "args": args},
+	fn := resume.MakePlainToolCall(
+		t.log,
 		t.TrajectoryID,
 		t.TenantID,
+		stepN,
 		seq,
-	); err != nil {
-		return nil, err
-	}
-	if _, err := t.log.Append(
-		"TOOL_RESULT",
-		&step,
-		map[string]any{"result": result},
-		t.TrajectoryID,
-		t.TenantID,
-		seq+1,
-	); err != nil {
+		tool.Name,
+		tool.Fn,
+	)
+	result, err := fn(args)
+	if err != nil {
 		return nil, err
 	}
 	return &ToolResult{StepN: stepN, Result: result}, nil
--- a/go/trajir/client/client_test.go
+++ b/go/trajir/client/client_test.go
@@ -2,6 +2,7 @@ package client_test
 
 import (
 	"context"
+	"errors"
 	"strings"
 	"sync/atomic"
 	"testing"
@@ -197,5 +198,32 @@ func TestSandboxRejectsNonIdempotent(t *testing.T) {
 	}
 	if _, err := tr.ExecTool(1, 2, tool, nil); err == nil {
 		t.Fatal("expected sandbox reject")
+	}
+}
+
+func TestExecToolLogsToolCallOnError(t *testing.T) {
+	dir := t.TempDir()
+	tr, err := client.OpenTrajectory("demo", "t-err", client.Options{WorkDir: dir})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tr.Close()
+	boom := errors.New("tool fail")
+	tool := resume.Tool{
+		Name:   "fetch",
+		Effect: effects.READ_ONLY,
+		Fn:     func(args map[string]any) (any, error) { return nil, boom },
+	}
+	_, err = tr.ExecTool(1, 2, tool, map[string]any{"url": "https://example.com"})
+	if !errors.Is(err, boom) {
+		t.Fatalf("err=%v want %v", err, boom)
+	}
+	ok, err := tr.Log().Has("t-err", "demo", 1, "TOOL_CALL", intPtr(2))
+	if err != nil || !ok {
+		t.Fatalf("TOOL_CALL has=%v err=%v", ok, err)
+	}
+	hasResult, err := tr.Log().Has("t-err", "demo", 1, "TOOL_RESULT", intPtr(3))
+	if err != nil || hasResult {
+		t.Fatalf("expected no TOOL_RESULT on error, got %v (err: %v)", hasResult, err)
 	}
 }
--- a/go/trajir/resume/gate.go
+++ b/go/trajir/resume/gate.go
@@ -87,3 +87,51 @@ func MakeGatedToolCall(
 		return result, nil
 	}
 }
+
+// MakePlainToolCall wraps toolFn for non-gated tools with audit logging.
+//
+// Seq layout matches Python: TOOL_CALL at seq, TOOL_RESULT at seq+1.
+// TOOL_CALL is recorded prior to tool execution so failed attempts are audited.
+// On success, TOOL_RESULT is recorded at seq+1 with {"result": ...}.
+func MakePlainToolCall(
+	log *nodelog.NodeLog,
+	trajectoryID, tenantID string,
+	stepN, seq int,
+	toolName string,
+	toolFn ToolFunc,
+) ToolFunc {
+	return func(args map[string]any) (any, error) {
+		if args == nil {
+			args = map[string]any{}
+		}
+		step := stepN
+
+		if _, err := log.Append(
+			"TOOL_CALL",
+			&step,
+			map[string]any{"tool": toolName, "args": args},
+			trajectoryID,
+			tenantID,
+			seq,
+		); err != nil {
+			return nil, err
+		}
+
+		result, err := toolFn(args)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, err := log.Append(
+			"TOOL_RESULT",
+			&step,
+			map[string]any{"result": result},
+			trajectoryID,
+			tenantID,
+			seq+1,
+		); err != nil {
+			return nil, err
+		}
+		return result, nil
+	}
+}
--- a/go/trajir/resume/step.go
+++ b/go/trajir/resume/step.go
@@ -135,37 +135,25 @@ func RunStep(
 			result, err = durable.Tool(ctx, cfg.Backend, cfg.WorkflowID, stepKey, func(context.Context) (any, error) {
 				return gated(args)
 			})
-		} else {
-			fn := tool.Fn
-			result, err = durable.Tool(ctx, cfg.Backend, cfg.WorkflowID, stepKey, func(context.Context) (any, error) {
-				return fn(args)
-			})
-			if err == nil {
-				// Plain tools still need IR history for audit; gate path already logs.
-				if _, err := cfg.Log.Append(
-					"TOOL_CALL",
-					&step,
-					map[string]any{"tool": call.Name, "args": args},
-					cfg.TrajectoryID,
-					cfg.TenantID,
-					seq,
-				); err != nil {
-					return nil, err
-				}
-				if _, err := cfg.Log.Append(
-					"TOOL_RESULT",
-					&step,
-					map[string]any{"result": result},
-					cfg.TrajectoryID,
-					cfg.TenantID,
-					seq+1,
-				); err != nil {
-					return nil, err
-				}
+			if err != nil {
+				return nil, err
 			}
-		}
-		if err != nil {
-			return nil, err
+		} else {
+			plain := MakePlainToolCall(
+				cfg.Log,
+				cfg.TrajectoryID,
+				cfg.TenantID,
+				stepN,
+				seq,
+				call.Name,
+				tool.Fn,
+			)
+			result, err = durable.Tool(ctx, cfg.Backend, cfg.WorkflowID, stepKey, func(context.Context) (any, error) {
+				return plain(args)
+			})
+			if err != nil {
+				return nil, err
+			}
 		}
 		results = append(results, result)
 	}
--- a/go/trajir/resume/step_test.go
+++ b/go/trajir/resume/step_test.go
@@ -309,7 +309,7 @@ func TestRunStepSandboxModeBlocksNonIdempotentWrite(t *testing.T) {
 	}
 }
 
-func TestRunStepNonGatedToolErrorNotLogged(t *testing.T) {
+func TestRunStepNonGatedToolErrorLogsToolCall(t *testing.T) {
 	ctx := context.Background()
 	nl, backend := testEnv(t)
 	boom := errors.New("tool boom")
@@ -335,7 +335,120 @@ func TestRunStepNonGatedToolErrorNotLogged(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if ok {
-		t.Fatal("TOOL_CALL should not be logged when non-gated tool errors")
+	if !ok {
+		t.Fatal("TOOL_CALL must be logged when non-gated tool errors")
+	}
+	hasResult, err := nl.Has("t1", "demo", 1, "TOOL_RESULT", intPtr(3))
+	if err != nil || hasResult {
+		t.Fatalf("expected no TOOL_RESULT on error, got %v (err: %v)", hasResult, err)
+	}
+}
+
+func TestRunStepReplayDoesNotDuplicateToolNodes(t *testing.T) {
+	ctx := context.Background()
+	nl, backend := testEnv(t)
+	var executions atomic.Int32
+
+	cfg := resume.RunStepConfig{
+		Log:          nl,
+		Backend:      backend,
+		TenantID:     "demo",
+		TrajectoryID: "t1",
+		WorkflowID:   "wf-replay-nodup",
+		Tools: map[string]resume.Tool{
+			"echo": {
+				Name:   "echo",
+				Effect: effects.PURE,
+				Fn: func(args map[string]any) (any, error) {
+					executions.Add(1)
+					return args["msg"], nil
+				},
+			},
+		},
+	}
+	model := func(context.Context, map[string]any) (map[string]any, error) {
+		return map[string]any{
+			"tool_calls": []any{
+				map[string]any{"name": "echo", "args": map[string]any{"msg": "hi"}},
+			},
+		}, nil
+	}
+
+	if _, err := resume.RunStep(ctx, cfg, 1, model, map[string]any{}); err != nil {
+		t.Fatal(err)
+	}
+	if executions.Load() != 1 {
+		t.Fatalf("executions=%d want 1", executions.Load())
+	}
+
+	// Replay the step under the same workflow ID.
+	if _, err := resume.RunStep(ctx, cfg, 1, model, map[string]any{}); err != nil {
+		t.Fatal(err)
+	}
+	if executions.Load() != 1 {
+		t.Fatalf("executions after replay=%d want 1", executions.Load())
+	}
+
+	nodes, err := nl.ListNodes("t1", "demo")
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Expected exactly: PROJECT_CONTEXT, DECISION, TOOL_CALL, TOOL_RESULT, COMMIT_STEP
+	if len(nodes) != 5 {
+		t.Fatalf("expected 5 nodes after replay, got %d", len(nodes))
+	}
+}
+
+func TestRunStepReplayAfterToolFailureDoesNotConflict(t *testing.T) {
+	ctx := context.Background()
+	nl, backend := testEnv(t)
+	var calls atomic.Int32
+	boom1 := errors.New("timeout after 100ms")
+	boom2 := errors.New("timeout after 150ms")
+
+	cfg := resume.RunStepConfig{
+		Log:          nl,
+		Backend:      backend,
+		TenantID:     "demo",
+		TrajectoryID: "t1",
+		WorkflowID:   "wf-fail-replay",
+		Tools: map[string]resume.Tool{
+			"failing": {
+				Name:   "failing",
+				Effect: effects.PURE,
+				Fn: func(map[string]any) (any, error) {
+					c := calls.Add(1)
+					if c == 1 {
+						return nil, boom1
+					}
+					return nil, boom2
+				},
+			},
+		},
+	}
+	model := func(context.Context, map[string]any) (map[string]any, error) {
+		return map[string]any{
+			"tool_calls": []any{
+				map[string]any{"name": "failing", "args": map[string]any{}},
+			},
+		}, nil
+	}
+
+	if _, err := resume.RunStep(ctx, cfg, 1, model, nil); !errors.Is(err, boom1) {
+		t.Fatalf("expected boom1, got %v", err)
+	}
+
+	// Retry with differently-worded error: must surface boom2 without slot conflict.
+	if _, err := resume.RunStep(ctx, cfg, 1, model, nil); !errors.Is(err, boom2) {
+		t.Fatalf("expected boom2 on replay, got %v", err)
+	}
+
+	nodes, err := nl.ListNodes("t1", "demo")
+	if err != nil {
+		t.Fatal(err)
+	}
+	// PROJECT_CONTEXT, DECISION, TOOL_CALL (no TOOL_RESULT on error)
+	if len(nodes) != 3 {
+		t.Fatalf("expected 3 nodes, got %d", len(nodes))
 	}
 }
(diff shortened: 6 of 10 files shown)


Ayush4958 (changes requested):
This implementation creates a severe architectural regression and misses part of the forensic context.

1. **Performance Regression (Fast-Path Destroyed):** By moving `t.log.Append("TOOL_CALL")` *before* the tool execution, you are injecting a synchronous, blocking database write into the critical path of every non-gated tool. This completely destroys the performance optimization that makes non-gated tools (like pure math or read-only API calls) useful in the first place.

2. **Missing Forensic Context:** When the tool errors, this code drops the error on the floor without logging a `TOOL_RESULT`. The durable log will show that a tool was attempted, but give zero context on what the actual error was.

**The Fix (Lazy-Log on Error):**
Please revert the pre-execution logging to restore the fast path. Instead, catch the error after execution. If an error occurs, write both `TOOL_CALL` and `TOOL_RESULT`right before returning.

Siddhartha-singh01 (commented):
Went through #265 against #262.

Bug is real. Plain tools used to call Fn first then append TOOL_CALL, so a failed Fn left NodeLog with no attempt recorded. Same hole in Go client, Go resume, Python client, and Python resume. Fix lines up across all four call sites.

Line notes are on the diff. Short take:

Happy path Append count did not grow. Order changed only. Error path now writes TOOL_CALL where it wrote nothing before, which is what #262 asked for.

Ayush's note about missing error context is fair. After a fail you get TOOL_CALL with args and zero detail on the error itself. #262 also allowed recording failure details. Tests currently lock in no TOOL_RESULT on failure. Worth deciding whether this PR should also append a TOOL_RESULT (or sibling node) with the error string on the fail branch in both SDKs, or leave that as a follow up. Lazy log of both nodes only on error is another shape and keeps the old success order.

On the fast path worry: success still does the same two Appends as before. You are not adding a new write on the happy path, only moving TOOL_CALL earlier. The new write is on the error path, and that is the point of the bugfix.

CI green, DCO signed, scope stays on the audit gap.

Leaving as COMMENT while the TOOL_RESULT on error question gets settled with Ayush.

go/trajir/client/client.go
```diff
@@ -241,6 +237,10 @@ func (t *Trajectory) ExecTool(stepN, seq int, tool resume.Tool, args map[string]
 	); err != nil {
 		return nil, err
 	}
+	result, err := tool.Fn(args)
```
Siddhartha-singh01: TOOL_CALL now lands before Fn. On error you return here with no TOOL_RESULT. Matches the tests as written. If we want the error string in NodeLog, this is the spot to Append TOOL_RESULT (or similar) with err before return.

go/trajir/resume/step.go
```diff
@@ -136,22 +136,21 @@ func RunStep(
 				return gated(args)
 			})
 		} else {
+			if _, err := cfg.Log.Append(
```
Siddhartha-singh01: Moved Append ahead of durable.Tool in the plain tool branch. Same contract as ExecTool. If Append fails you never run the tool, which is fine.

go/trajir/client/client_test.go
```diff
@@ -199,3 +200,33 @@ func TestSandboxRejectsNonIdempotent(t *testing.T) {
...
+	if err != nil || !ok {
+		t.Fatalf("TOOL_CALL has=%v err=%v", ok, err)
+	}
+	hasResult, err := tr.Log().Has("t-err", "demo", 1, "TOOL_RESULT", intPtr(3))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if hasResult {
+		t.Fatal("TOOL_RESULT must not be logged when tool fails")
+	}
```
Siddhartha-singh01, line 228: Locks in TOOL_RESULT must not exist when the tool fails. Fine for the current design. Flip this assert if we add error payload logging.

go/trajir/resume/step_test.go
```diff
@@ -309,7 +309,7 @@ func TestRunStepSandboxModeBlocksNonIdempotentWrite(t *testing.T) {
 	}
 }
 
-func TestRunStepNonGatedToolErrorNotLogged(t *testing.T) {
+func TestRunStepNonGatedToolErrorLogsToolCall(t *testing.T) {
```
Siddhartha-singh01, line 312: Rename matches the new behavior. Old name said NotLogged, which was the pre fix contract.

go/trajir/resume/step_test.go
```diff
@@ -335,7 +335,14 @@ func TestRunStepNonGatedToolErrorNotLogged(t *testing.T) {
...
-		t.Fatal("TOOL_CALL should not be logged when non-gated tool errors")
+	if !ok {
+		t.Fatal("TOOL_CALL must be logged when non-gated tool errors")
+	}
+	hasResult, err := nl.Has("t1", "demo", 1, "TOOL_RESULT", intPtr(3))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if hasResult {
+		t.Fatal("TOOL_RESULT must not be logged when non-gated tool errors")
```
Siddhartha-singh01: Same no TOOL_RESULT on error rule as the client test. Keep Go and Python aligned if that rule changes.

client/python/trajectory_client.py
```diff
@@ -125,6 +122,7 @@ def exec_tool(trajectory: Trajectory, step_n: int, call: dict, tool: Tool, seq:
             trajectory.tenant_id,
             seq=seq,
         )
+        result = tool.fn(**call["args"])
```
Siddhartha-singh01: Same reorder as Go ExecTool. Old comment about claim gating dropped with the move. Behavior matches the Go client.

pkg/trajectory_ir/resume/step.py
```diff
@@ -150,6 +147,7 @@ def run_step(step_n: int, model_call, context: dict):
                     tenant_id,
                     seq=seq,
                 )
+                result = durable_tool(tool.fn)(**call["args"])
```
Siddhartha-singh01: Same as the Go resume else branch. R03 note went away with the move. Good that Python and Go stay in lockstep here.

test/unit/test_client_sdk_plain_tool_logging.py
```diff
@@ -34,3 +36,20 @@ def test_exec_tool_logs_two_plain_tools_same_step_distinct_seq(tmp_path, monkeyp
...
+    def failing_fn(x):
+        raise ValueError("boom")
+
+    tool = Tool(name="fail", fn=failing_fn, effect_class=EffectClass.PURE)
+    with pytest.raises(ValueError, match="boom"):
+        exec_tool(traj, step_n=1, call={"args": {"x": 1}}, tool=tool, seq=2)
+
+    log = NodeLog(db_path)
+    assert log.has("plain-err", "demo", 1, "TOOL_CALL", seq=2)
+    assert not log.has("plain-err", "demo", 1, "TOOL_RESULT", seq=3)
```
Siddhartha-singh01, line 55: Solid coverage for the Python client path. Call present at seq 2, result absent at seq 3.

test/unit/test_step.py
```diff
@@ -45,3 +46,33 @@ def test_plain_tool_logs_tool_call_and_result(tmp_path, monkeypatch):
...
+    }
+
+    run_step = make_run_step(node_log, TENANT_ID, "test-step-fail", tool_registry)
+    init_backend(app_name="test-step-fail")
+
+    with SetWorkflowID("test-step-fail"), pytest.raises(RuntimeError, match="tool error"):
+        run_step(step_n=1, model_call=_model_call_fail, context={})
+
+    assert node_log.has("test-step-fail", TENANT_ID, 1, "TOOL_CALL", seq=2)
+    assert not node_log.has("test-step-fail", TENANT_ID, 1, "TOOL_RESULT", seq=3)
```
Siddhartha-singh01, line 91: Covers the DBOS resume path with a failing READ_ONLY tool. Same contract as the Go resume test.

Ayush4958 (changes requested):
**Finding 1 (BLOCKING) - Replay Duplication in `step.go` and `step.py`:**
In `step.go`, `TOOL_CALL` is now logged **outside and before** `durable.Tool(...)`. Since `durable.Tool` is the DBOS replay wrapper, on every workflow resume/retry, `TOOL_CALL` gets appended again. This creates **phantom duplicate entries** in the audit log on any resume — which is the exact scenario this whole system exists to handle correctly.

**Finding 2 (HIGH) - Error context still missing:**
A `TOOL_CALL` with no `TOOL_RESULT` on failure tells investigators nothing. The correct fix is lazy-logging: on error, write both `TOOL_CALL` AND a `TOOL_RESULT` containing the error message before returning.

**Finding 3 - Test coverage gap in `step_test.go`:**
The Go step test only asserts `TOOL_CALL` is present. It doesn't assert `TOOL_RESULT` is absent. The client test does both. They should match.

go/trajir/resume/gate.go
```diff
@@ -87,3 +87,70 @@ func MakeGatedToolCall(
...
+			map[string]any{"tool": toolName, "args": args},
+			trajectoryID,
+			tenantID,
+			seq,
+		); logErr != nil {
+			return nil, logErr
+		}
+		if _, logErr := log.Append(
+			"TOOL_RESULT",
+			&step,
```
jakharmonika364: `TOOL_RESULT` is appended inside the closure `durable.Step` only memoizes on success. if a non-gated tool fails once, its error gets logged at this (step, seq) slot but the step stays unmemoized, so a resumed retry (explicitly expected per R03 for PURE/READ_ONLY tools) re-runs the tool. A successful retry then tries to append a different-hash result to the same slot and hits `ErrSlotConflict` permanently. since the log is append-only/content- addressed. The trajectory can never complete after one transient failure. Same defect in `gate.py:122`.
Need to either memoize the error case too, or route the error result through a path that doesn't collide with a later successful append at the same slot.

go/trajir/resume/gate.go
```diff
@@ -87,3 +87,70 @@ func MakeGatedToolCall(
...
+//
+// Seq layout matches Python: TOOL_CALL at seq, TOOL_RESULT at seq+1.
+// On success, it records TOOL_CALL and TOOL_RESULT with {"result": ...}.
+// On error, it records TOOL_CALL and TOOL_RESULT with {"error": ...} before returning the error.
+func MakePlainToolCall(
+	log *nodelog.NodeLog,
+	trajectoryID, tenantID string,
+	stepN, seq int,
+	toolName string,
+	toolFn ToolFunc,
```
jakharmonika364, line 101: the error and success branches duplicate the full `TOOL_CALL`/`TOOL_RESULT` append sequence. Worth factoring into one shared "compute outcome, then append once" path, the duplicaion here likely hid the slot-conflict issue above during review

jakharmonika364 (changes requested):

Ayush4958 (dismissed):
In `step.py`, the exception reconstruction via `getattr(builtins, exc_type_name)` should be wrapped in a try/except to handle exceptions with non-standard constructors :-

```python
try:
    raise exc_cls(err_msg)
except TypeError:
    raise RuntimeError(err_msg)
```

It's not a blocker, 
PR is solid Otherwise 

LGTM

Ayush4958:
cc: @jakharmonika364 @Siddhartha-singh01 

further approval can merge this PR

go/trajir/resume/step.go
```diff
@@ -135,37 +136,43 @@ func RunStep(
...
+				cfg.Log,
+				cfg.TrajectoryID,
+				cfg.TenantID,
+				stepN,
+				seq,
+				call.Name,
+				tool.Fn,
+			)
+			var lastErr error
+			rawOutcome, stepErr := durable.Tool(ctx, cfg.Backend, cfg.WorkflowID, stepKey, func(context.Context) (any, error) {
```
jakharmonika364: This narrows the retry-conflict bug but doesn't fix it, log writes still happen before the memo save, so a crash in that window causes the same conflict on replay. Same on the Python side (step.py:163). Think the log write needs to move after memoization, not just after-the-fact retry handling.

pkg/trajectory_ir/resume/step.py
```diff
@@ -139,25 +151,25 @@ def run_step(step_n: int, model_call, context: dict):
...
+                    seq,
+                    call["name"],
+                    tool.fn,
                 )
+                wrapped = _wrap_durable_plain(plain)
+                outcome = durable_tool(wrapped)(**call["args"])
+                if outcome.get("__failed"):
+                    err_msg = outcome.get("error", "tool failed")
+                    exc_type_name = outcome.get("__exc_type", "")
+                    exc_cls = getattr(builtins, exc_type_name, None)
```
jakharmonika364: Exception reconstruction only looks in `builtins`, so any custom exception (e.g. `ToolExecutionError`) gets replayed as a plain `RuntimeError`. Breaks callers doing except `ToolExecutionError:`.

go/trajir/resume/step.go
```diff
@@ -135,37 +136,43 @@ func RunStep(
...
+			})
+			if stepErr != nil {
+				return nil, stepErr
+			}
+			outcome, err := asStringMap(rawOutcome)
+			if err != nil {
+				return nil, fmt.Errorf("resume: tool outcome: %w", err)
+			}
+			if failed, _ := outcome["__failed"].(bool); failed {
+				if lastErr != nil {
```
jakharmonika364: On replay, if the failure comes from the memo instead of a fresh run, `lastErr` is nil and we fall back to `errors.New(msg)`, loses the original error type, breaks `errors.Is`/`errors.As` after replay.

jakharmonika364 (changes requested):
same root cause as before, just a narrower window now. Thanks! + two new error-identity bugs from the fix itself.

pkg/trajectory_ir/resume/gate.py
```diff
@@ -79,3 +79,45 @@ def gated(**kwargs):
...
+
+    def plain(**kwargs):
+        err: Exception | None = None
+        result = None
+        try:
+            result = tool_fn(**kwargs)
+            payload = {"result": result}
+        except Exception as e:
+            err = e
+            payload = {"error": str(e)}
```
jakharmonika364: Logging `str(e)` into the `TOOL_RESULT` payload means a retry with differently-worded error text (timing, connection info) hits a slot conflict instead of the real error. Repro'd locally: second failure raises `SlotConflictError` instead of the actual exception.

go/trajir/resume/gate.go
```diff
@@ -87,3 +87,55 @@ func MakeGatedToolCall(
...
+	return func(args map[string]any) (any, error) {
+		if args == nil {
+			args = map[string]any{}
+		}
+		step := stepN
+
+		result, err := toolFn(args)
+		payload := map[string]any{"result": result}
+		if err != nil {
+			payload = map[string]any{"error": err.Error()}
```
jakharmonika364: Same issue as gate.py: `err.Error()` in the `TOOL_RESULT` payload means a retried failure with different error text conflicts at the same slot instead of surfacing the real error.

jakharmonika364 (changes requested):

go/trajir/resume/gate.go
```diff
@@ -87,3 +87,51 @@ func MakeGatedToolCall(
...
+	toolFn ToolFunc,
+) ToolFunc {
+	return func(args map[string]any) (any, error) {
+		if args == nil {
+			args = map[string]any{}
+		}
+		step := stepN
+
+		if _, err := log.Append(
+			"TOOL_CALL",
```
Siddhartha-singh01, line 110: TOOL_CALL before toolFn. Same args on retry hit the same node id, so Append is a no op. That is what keeps durable retries from duplicating the attempt row.

go/trajir/resume/gate.go
```diff
@@ -87,3 +87,51 @@ func MakeGatedToolCall(
...
+			&step,
+			map[string]any{"tool": toolName, "args": args},
+			trajectoryID,
+			tenantID,
+			seq,
+		); err != nil {
+			return nil, err
+		}
+
+		result, err := toolFn(args)
```
Siddhartha-singh01, line 120: On err you return with no TOOL_RESULT. Matches the tests and avoids putting a moving error string into a content addressed slot. Audit still has the attempt; it just has no failure detail.

go/trajir/resume/step.go
```diff
@@ -135,38 +135,26 @@ func RunStep(
...
+			plain := MakePlainToolCall(
+				cfg.Log,
+				cfg.TrajectoryID,
+				cfg.TenantID,
+				stepN,
+				seq,
+				call.Name,
+				tool.Fn,
+			)
 			result, err = durable.Tool(ctx, cfg.Backend, cfg.WorkflowID, stepKey, func(context.Context) (any, error) {
```
Siddhartha-singh01, line 151: Right place for the wrapper. Logging rides inside durable.Tool, so a successful memo replay never re enters MakePlainToolCall. Crash after TOOL_RESULT and before Save still re runs the tool on resume, which is fine for PURE.

pkg/trajectory_ir/resume/gate.py
```diff
@@ -79,3 +79,36 @@ def gated(**kwargs):
...
+    """Wrap a non-gated tool with audit logging.
+
+    Occupies two seq slots: TOOL_CALL at `seq`, and TOOL_RESULT at `seq + 1`.
+    TOOL_CALL is recorded prior to tool execution so failed attempts are audited.
+    On success, TOOL_RESULT is recorded at seq + 1 with {"result": ...}.
+    """
+
+    def plain(**kwargs):
+        node_log.append(
+            "TOOL_CALL",
```
Siddhartha-singh01, line 94: Python mirror of the Go plain wrapper. Good. Same contract: call first, result only on success.

pkg/trajectory_ir/resume/step.py
```diff
@@ -139,25 +139,16 @@ def run_step(step_n: int, model_call, context: dict):
...
-                    seq=seq,
-                )
-                node_log.append(
-                    "TOOL_RESULT",
                     step_n,
-                    {"result": result},
-                    trajectory_id,
-                    tenant_id,
-                    seq=seq + 1,
+                    seq,
```
Siddhartha-singh01, line 147: plain wrapped then durable_tool(plain). Matches Go. This is the fix for the early revision that logged outside durable and duplicated on resume.

go/trajir/client/client.go
```diff
@@ -227,28 +226,17 @@ func (t *Trajectory) ExecTool(stepN, seq int, tool resume.Tool, args map[string]
...
 	}
-	result, err := tool.Fn(args)
-	if err != nil {
-		return nil, err
-	}
-	if _, err := t.log.Append(
-		"TOOL_CALL",
-		&step,
-		map[string]any{"tool": tool.Name, "args": args},
+	fn := resume.MakePlainToolCall(
```
Siddhartha-singh01, line 229: Client path has no durable memo, which is fine for ExecTool. Same MakePlainToolCall helper keeps client and resume aligned.

client/python/trajectory_client.py
```diff
@@ -114,25 +114,16 @@ def exec_tool(trajectory: Trajectory, step_n: int, call: dict, tool: Tool, seq:
...
         result = fn(**call["args"])
     else:
-        # Not claim-gated, but still needs IR history audit; gate path
-        # already logs its own TOOL_CALL/TOOL_RESULT nodes.
-        result = tool.fn(**call["args"])
-        log.append(
-            "TOOL_CALL",
-            step_n,
-            {"tool": tool.name, "args": call["args"]},
+        fn = make_plain_tool_call(
```
Siddhartha-singh01, line 117: Same client side reorder via make_plain_tool_call. Consistent with Go ExecTool.

go/trajir/resume/step_test.go
```diff
@@ -335,7 +335,120 @@ func TestRunStepNonGatedToolErrorNotLogged(t *testing.T) {
...
+			},
+		}, nil
+	}
+
+	if _, err := resume.RunStep(ctx, cfg, 1, model, nil); !errors.Is(err, boom1) {
+		t.Fatalf("expected boom1, got %v", err)
+	}
+
+	// Retry with differently-worded error: must surface boom2 without slot conflict.
+	if _, err := resume.RunStep(ctx, cfg, 1, model, nil); !errors.Is(err, boom2) {
```
Siddhartha-singh01, line 442: This asserts failure is NOT memoized: second RunStep must re run and return boom2. Compare with the Python twin, which expects the first error text again. Need one shared story for failure replay.

test/unit/test_step.py
```diff
@@ -45,3 +46,117 @@ def test_plain_tool_logs_tool_call_and_result(tmp_path, monkeypatch):
...
+    }
+
+    run_step = make_run_step(node_log, TENANT_ID, traj_id, tool_registry)
+    init_backend(app_name="test-step-fail-replay")
+
+    with SetWorkflowID(traj_id), pytest.raises(CustomToolError, match="timeout after 100ms"):
+        run_step(step_n=1, model_call=_model_call_replay_fail, context={})
+
+    # Replay under same workflow ID: returns recorded exception without slot conflict
+    with SetWorkflowID(traj_id), pytest.raises(CustomToolError, match="timeout after 100ms"):
```
Siddhartha-singh01: Second raise still matches timeout after 100ms, so this assumes DBOS memoizes the first failure. Go test expects the opposite. Align these before calling the failure path done.

go/trajir/client/client_test.go
```diff
@@ -199,3 +200,30 @@ func TestSandboxRejectsNonIdempotent(t *testing.T) {
 		t.Fatal("expected sandbox reject")
 	}
 }
+
+func TestExecToolLogsToolCallOnError(t *testing.T) {
```
Siddhartha-singh01, line 204: Covers the #262 hole on the client path. TOOL_CALL present, TOOL_RESULT absent. Good.

test/unit/test_client_sdk_plain_tool_logging.py
```diff
@@ -34,3 +36,20 @@ def test_exec_tool_logs_two_plain_tools_same_step_distinct_seq(tmp_path, monkeyp
     assert log.has("plain-2", "demo", 1, "TOOL_RESULT", seq=3)
     assert log.has("plain-2", "demo", 1, "TOOL_CALL", seq=4)
     assert log.has("plain-2", "demo", 1, "TOOL_RESULT", seq=5)
+
+
+def test_exec_tool_logs_tool_call_when_tool_raises(tmp_path, monkeypatch):
```
Siddhartha-singh01, line 41: Python client twin of the Go ExecTool error test. Solid.

Siddhartha-singh01 (commented):
Went through tip b358980 again against #262 and the earlier review thread.

Shape is much clearer now. Logging lives inside MakePlainToolCall / make_plain_tool_call, and that wrapper is what durable.Tool runs. Success replay stays inside the memo so you do not get phantom TOOL_CALL rows. Same content Append is idempotent, so a retried TOOL_CALL with the same args is a no op. Skipping TOOL_RESULT on failure avoids stuffing a changing error string into a content addressed slot, which is what blew up the middle revisions.

#262 goal is met: failed plain tools leave a TOOL_CALL.

Open points I still want eyes on before merge:

1. Forensic gap is intentional in this tip. Failed attempts get TOOL_CALL only. No error payload in NodeLog. Fine if that is the agreed R03 tradeoff. Call it out in the PR summary so nobody thinks it is unfinished.
2. Go and Python failure replay tests disagree. Go expects the second RunStep to re execute and surface boom2. Python expects the second run_step to raise the first CustomToolError text again. One of those is the real durable contract; the other should match it.
3. Monika earlier notes about logging err text into TOOL_RESULT and builtins exception rebuild look fixed or removed on this tip. Worth confirming those review threads are resolved on the current files, not an older push.

CI is green. Leaving COMMENT only.

test/unit/test_step.py
```diff
@@ -45,3 +59,125 @@ def test_plain_tool_logs_tool_call_and_result(tmp_path, monkeypatch):
...
+
+def _model_call_replay_fail(context):
+    return {"tool_calls": [{"name": "fail", "args": {}}]}
+
+
+def test_plain_tool_failure_replay_does_not_conflict(tmp_path, monkeypatch):
+    monkeypatch.chdir(tmp_path)
+    node_log = NodeLog(str(tmp_path / "nodes_fail_replay.sqlite"))
+    traj_id = "test-step-fail-replay"
+    _REPLAY_FAIL_CALLS["n"] = 0
```
Siddhartha-singh01, line 157: This was the mismatch I flagged. Switching to restate local_memo and expecting timeout after 150ms on retry matches Go. Local memo only stores success, so failures re execute. Nice fix.

go/trajir/resume/step_test.go
Siddhartha-singh01, line 442: Still the right contract: second RunStep surfaces boom2, three nodes, no TOOL_RESULT. Python now agrees.

go/trajir/resume/gate.go
Siddhartha-singh01, line 120: Returning without TOOL_RESULT on err is the right tradeoff for R03 retries with changing messages. Documented in the commit body now.

pkg/trajectory_ir/resume/gate.py
```diff
@@ -79,3 +79,36 @@ def gated(**kwargs):
...
+            step_n,
+            {"tool": tool_name, "args": kwargs},
+            trajectory_id,
+            tenant_id,
+            seq=seq,
+        )
+
+        result = tool_fn(**kwargs)
```
Siddhartha-singh01, line 102: Same success only TOOL_RESULT path as Go. Good lockstep.

go/trajir/resume/step.go
Siddhartha-singh01, line 151: Wrapper inside durable.Tool is still the important structural fix from the early revisions.

Siddhartha-singh01 (approved):
Re reviewed tip c6fa9a0.

The change since my last pass fixes the Go vs Python failure replay mismatch. Python now uses the restate local_memo hooks (success only memo, same idea as Go durable.Tool) and expects the second error text on retry. Commit message also spells out why TOOL_RESULT is omitted on failure.

Still good on the core design:
MakePlainToolCall / make_plain_tool_call logs TOOL_CALL before Fn, TOOL_RESULT only on success, wrapper runs inside durable so success replay does not duplicate nodes, same content Append stays idempotent.

#262 is covered. CI looked green aside from still running Analyze/Go when I checked.

Monika had older CHANGES_REQUESTED on intermediate pushes (error string in TOOL_RESULT, crash window). Current tip does not put error text in TOOL_RESULT. Worth her resolving those threads, but from my side this is good to merge.

LGTM.

jakharmonika364 (approved):
Logging now lives inside the durable wrapper, so replay doesn't duplicate `TOOL_CALL`, and dropping `TOOL_RESULT` on error kills the slot-conflict class instead of narrowing it again. Threads from my last rounds look resolved on this tip. LGTM.

Ayush-Patel-56 (approved):
ahh! too long PR

Cheers!

Issue #197, opened 2026-08-14
ci: clear Scorecard pinned-deps and release token findings (Security tab 27)


Siddhartha-singh01:
## Why this is next (step 1)

GitHub Security shows ~27 **code scanning** alerts. They are almost all **OpenSSF Scorecard**, not Dependabot CVEs (Dependabot open count is 0).

## Goal

Land CI/workflow fixes so Scorecard **Pinned-Dependencies** and **Token-Permissions** drop after the next Scorecard run on `main`.

## Work already prepared (local)

Branch: `ci/scorecard-security-fixes` (commit `d93e1c1`)

- Hashed pip locks under `requirements/`
- `--require-hashes` + `--no-deps -e .` install pattern
- `govulncheck@v1.1.4`, gitleaks checksum verify
- syft via SHA-pinned `anchore/sbom-action`
- release.yml top-level `contents: read`, job-level write
- Docs in `docs/CI_HARDENING.md`

## Blocker to push

GitHub rejects workflow file updates without **`workflow`** OAuth scope:

```text
refusing to allow an OAuth App to create or update workflow without workflow scope
```

Maintainer must run once:

```powershell
gh auth refresh -h github.com -s workflow,repo
cd H:\Trajectory-IR
git checkout ci/scorecard-security-fixes
git push -u origin HEAD
gh pr create --title "ci: clear Scorecard pinned-deps and release token findings" --body "Closes this issue when merged."
```

## Acceptance

- [ ] PR opened from `ci/scorecard-security-fixes` (or equivalent)
- [ ] CI green
- [ ] Merged to `main`
- [ ] Scorecard re-run; pin + token alerts reduced
- [ ] Process-only checks remain documented (Maintained age, Code-Review, CII badge, Fuzzing, SAST history)

## Out of scope

- Full OSS-Fuzz
- Completing OpenSSF Best Practices badge
- Waiting 90 days for Maintained score

Siddhartha-singh01:
Still blocked on push: GitHub wants the workflow OAuth scope for updating .github/workflows/*.

Local branch ci/scorecard-security-fixes is rebased on current main (commit d93be4a).

When you have a minute on your machine:

```
gh auth refresh -h github.com -s workflow,repo
cd H:\Trajectory-IR
git checkout ci/scorecard-security-fixes
git push -u origin HEAD
gh pr create --title "ci: clear Scorecard pinned-deps and release token findings" --body "Closes #197."
```

[...]

Issue #179, opened 2026-08-13
feat(python): trajir-pkg-sig-v1 sign/verify parity + R09–R11 (Phase C)


Siddhartha-singh01:
## Parent

Part of epic [#149](https://github.com/Coder-s-OG-s/Trajectory-IR/issues/149) (package digital signatures).

| Phase | Tracker | Status |
|-------|---------|--------|
| A Spec | PR [#176](https://github.com/Coder-s-OG-s/Trajectory-IR/pull/176) | Merged (README §9.1) |
| B Go Sign/Verify | [#177](https://github.com/Coder-s-OG-s/Trajectory-IR/issues/177) / PR [#178](https://github.com/Coder-s-OG-s/Trajectory-IR/pull/178) | In progress / review |
| **C Python + conformance** | **This issue** | Open |
| D Sigstore optional | Later | Not opened |

**Depends on:** Phase B goldens and Go APIs (README §9.1). Prefer merge of #178 before landing Python, or develop against the same golden file.

## Scope

Implement **byte-identical** `trajir-pkg-sig-v1` in the Python reference SDK so a package signed in Go verifies in Python and vice versa.

### Implementation
- [ ] Payload helper `trajir-pkg-payload-v1` (sorted zip members except `SIGNATURE`; `uint64_be(len(N)) || N || SHA-256(B)`)
- [ ] Domain-separated Ed25519: `SHA-256(b"trajir-pkg-sig-v1" + b"\x00" + payload_hash_raw)`
- [ ] Write/read zip member `SIGNATURE` (JSON per README §9.1)
- [ ] Keep `manifest["signature"]` always `None` / null
- [ ] `sign_package(path, private_key, meta=...)` and `verify_package(path, *,require_signature, trusted_keys, trusted_key_ids)`
- [ ] Optional export sign hook (parity with Go `ExportOptions.SignKey`) if export API allows
- [ ] Load/import: if `SIGNATURE` present, verify (fail closed); unsigned remains default-OK
- [ ] Consume Go goldens: `go/trajir/tir/testdata/sig_v1/payload_golden.json`

### Conformance (proposed R09–R11)
- [ ] **R09** — Sign thin package; verify succeeds; mutate a covered member; verify fails
- [ ] **R10** — Unsigned still passes R05 path; strict `require_signature` rejects unsigned
- [ ] **R11** — Go-signed package verifies in Python **and** Python-signed verifies in Go (CI or dual-language test)

### Docs
[...]

PR #266, opened 2026-08-20
feat(python): trajir-pkg-sig-v1 sign and verify (Phase C)

Siddhartha-singh01:
## Summary

Starts #179 Phase C.

Python now has byte matched trajir-pkg-sig-v1 helpers against the Go golden file, plus sign_package / verify_package, optional export_tir(sign_key=...), and load checks SIGNATURE when it is present.

## What is in

- pkg/trajectory_ir/package/signature.py
- export_tir sign_key (64 byte key, wrong length fails closed)
- load_tir verifies SIGNATURE when present (tamper fails even on unverified load path for signature)
- unit tests including Go golden payload / domain / key id
- cryptography dependency

## Still open on #179

- Full conformance R09 R10 R11 jobs in CI
- Broader docs pass
- Security review before merge

## Test plan

- [x] pytest test/unit/test_package_signature.py
- [x] ruff check / format
- [ ] CI green

## Note

#197 Scorecard PR still needs workflow OAuth on the push token (cannot open that one from here yet).

### DIFF
--- a/pkg/trajectory_ir/package/signature.py
+++ b/pkg/trajectory_ir/package/signature.py
@@ -125,6 +125,8 @@ def _collect_zip_members(
     for info in infos:
         name = info.filename
         _safe_member_name(name)
+        # Declared size is advisory only: a malicious header can understate
+        # file_size. Cap the read so a lying local header cannot OOM us.
         if info.file_size > MAX_UNCOMPRESSED_ENTRY_BYTES:
             raise TirLimitError(
                 f"zip member {name!r} uncompressed size {info.file_size} "
@@ -134,9 +136,14 @@ def _collect_zip_members(
             raise TirLimitError(
                 f"zip total uncompressed size would exceed limit {MAX_TOTAL_UNCOMPRESSED_BYTES}"
             )
-        data = zf.read(name)
+        with zf.open(name) as fp:
+            data = fp.read(MAX_UNCOMPRESSED_ENTRY_BYTES + 1)
         if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
             raise TirLimitError(f"zip member {name!r} expanded beyond per-entry limit")
+        if len(data) > budget:
+            raise TirLimitError(
+                f"zip total uncompressed size would exceed limit {MAX_TOTAL_UNCOMPRESSED_BYTES}"
+            )
         budget -= len(data)
         if name == SIGNATURE_MEMBER:
             if signature is not None:
--- a/pkg/trajectory_ir/package/signature.py
+++ b/pkg/trajectory_ir/package/signature.py
@@ -89,12 +89,18 @@ def payload_hash_from_members(members: dict[str, bytes]) -> bytes:
 
 
 def _safe_member_name(name: str) -> None:
+    # Match tir._safe_zip_name / Go safeZipName so load and sign agree.
     if not name or name.startswith(("/", "\\")):
         raise TirSignatureError(f"unsafe zip path: {name!r}")
     parts = name.replace("\\", "/").split("/")
-    for p in parts:
-        if p in ("", ".", ".."):
-            raise TirSignatureError(f"unsafe zip path: {name!r}")
+    if parts and parts[0] == "":
+        raise TirSignatureError(f"unsafe zip path: {name!r}")
+    if ".." in parts:
+        raise TirSignatureError(f"unsafe zip path: {name!r}")
+    if name.startswith("artifacts/") and not (
+        name == "artifacts/" or name.startswith("artifacts/cas/")
+    ):
+        raise TirSignatureError(f"unexpected artifact path: {name!r}")
 
 
 def _signing_key_from_bytes(key: bytes) -> SigningKey:
@@ -132,12 +138,9 @@ def _collect_zip_members(
                 f"zip member {name!r} uncompressed size {info.file_size} "
                 f"exceeds limit {MAX_UNCOMPRESSED_ENTRY_BYTES}"
             )
-        if info.file_size > budget:
-            raise TirLimitError(
-                f"zip total uncompressed size would exceed limit {MAX_TOTAL_UNCOMPRESSED_BYTES}"
-            )
+        limit = min(MAX_UNCOMPRESSED_ENTRY_BYTES, budget)
         with zf.open(name) as fp:
-            data = fp.read(MAX_UNCOMPRESSED_ENTRY_BYTES + 1)
+            data = fp.read(limit + 1)
         if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
             raise TirLimitError(f"zip member {name!r} expanded beyond per-entry limit")
         if len(data) > budget:
@@ -293,7 +296,7 @@ def sign_package(
     meta = meta or SignerMeta()
     signed_at = meta.signed_at or datetime.now(UTC)
     if signed_at.tzinfo is None:
-        signed_at = signed_at.replace(tzinfo=UTC)
+        raise TirSignatureError("signed_at must be timezone aware")
     signer: dict[str, str] = {"key_id": key_id(pub)}
     if meta.id:
         signer["id"] = meta.id
--- a/pkg/trajectory_ir/package/signature.py
+++ b/pkg/trajectory_ir/package/signature.py
@@ -98,14 +98,16 @@ def _safe_member_name(name: str) -> None:
 
 
 def _signing_key_from_bytes(key: bytes) -> SigningKey:
-    if len(key) == _PRIVATE_KEY_SIZE:
-        return SigningKey(key[:_SEED_SIZE])
-    if len(key) == _SEED_SIZE:
-        return SigningKey(key)
-    raise TirSignatureError("invalid ed25519 private key size")
+    # Match Go Sign: public API accepts only the full 64 byte private key.
+    if len(key) != _PRIVATE_KEY_SIZE:
+        raise TirSignatureError("invalid ed25519 private key size")
+    return SigningKey(key[:_SEED_SIZE])
 
 
-def _read_zip_members(path: Path) -> tuple[dict[str, bytes], bytes | None]:
+def _collect_zip_members(
+    zf: zipfile.ZipFile,
+) -> tuple[dict[str, bytes], bytes | None]:
+    """Read members from an already open archive with tir.py size budgets."""
     # Local import avoids a cycle: tir.py imports this module at load time.
     from trajectory_ir.package.tir import (
         MAX_TOTAL_UNCOMPRESSED_BYTES,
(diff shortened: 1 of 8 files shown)


Siddhartha-singh01:
First cut of Phase C. Please review the crypto path carefully. Happy to add R09 R10 conformance files in a follow on PR.

Ayush4958 (changes requested):
there is a **critical OOM/Zip-Bomb vulnerability (CWE-400)** in `signature.py`.

In `_read_zip_members`, the code blindly reads every byte of every file into an unbounded memory dictionary (`data = zf.read(name)`). Because `verify_package` is called during `load_tir`, this bypasses the strict `MAX_TOTAL_UNCOMPRESSED_BYTES` budget enforced in `tir.py`. A malicious package with a fake signature and a 50GB junk file will cause an instant Out-Of-Memory crash.

**The Fix:**
You must stream the file contents into `hashlib` in chunks rather than storing them in memory.

Please refactor `payload_hash_from_members` to compute the hash in `O(1)` memory. Example:

Ayush-Patel-56:
CI is failing, please look into it @Siddhartha-singh01

Siddhartha-singh01:
CI failure was the Python license scan. cryptography reports Apache-2.0 OR BSD-3-Clause which is not on the allow only list, and I cannot push workflow edits without workflow OAuth. Switched the crypto dep to PyNaCl (Apache-2.0). Same tests still pass.

Siddhartha-singh01:
License scan still failed on cffi 2.1 (MIT-0). Pinned cffi below 2.1 so it stays MIT. No workflow change needed.

Siddhartha-singh01:
Follow up: cffi 2.1 reports MIT-0 and trips the license scan. Pinned cffi below 2.1 so it stays MIT.

Ayush4958 (changes requested):
The real issue is that `_read_zip_members` lacks the uncompressed size bounds that `tir.py` enforces, leaving the `verify_package` and `sign_package` public APIs vulnerable to Zip-Bomb OOM crashes.

Since `.tir` packages are designed to fit in memory, you don't need to stream the hash. You just need to enforce the same budget limits established in `tir.py`.

**The Fix:**
Import `MAX_TOTAL_UNCOMPRESSED_BYTES` and `TirLimitError` from `trajectory_ir.package.tir`. Inside `_read_zip_members`, keep a running tally of `data = zf.read(name)`. If the total uncompressed bytes exceed the limit, raise `TirLimitError`. 

This closes the DoS backdoor with just three lines of code and keeps the zip rewriting logic trivial.

Siddhartha-singh01:
@Ayush4958 fixed the zip bomb path you flagged.

`_read_zip_members` now enforces the same budgets as `tir.py` (entry count, per entry uncompressed size, and total uncompressed size) before keeping member bytes in memory. Local import to avoid the tir/signature cycle. Added a unit test that shrinks the limits and expects `TirLimitError`.

pkg/trajectory_ir/package/tir.py
```diff
@@ -446,12 +472,22 @@ def _load_tir_impl(path: str | Path, *, verify: bool) -> TirPackage:
...
             raise TirVerificationError("thin package must not embed artifact bytes")
 
+    # Signature check after hash/seal integrity (README 9.1 order).
+    # Present but invalid SIGNATURE fails even for load_tir_unverified (tamper).
+    sig_info = None
+    if has_signature:
+        try:
+            sig_info = verify_package(path)
+        except TirSignatureError as e:
+            raise TirVerificationError(str(e)) from e
```
jakharmonika364: `verify_package(path)` reopens the file from disk after the first `ZipFile` handle (used to read manifest/nodes/seals) is already closed. If the file changes between the two opens, u can return package content from the first read alongside a "signature valid" result computed on the second read's bytes, a TOCTOU gap (ig CWE-367). The Go implementation avoids this on purpose (see its comment about never re-opening `path`). Please verify against the same already-open zip handle used to read the package content.

pkg/trajectory_ir/package/signature.py
```diff
@@ -0,0 +1,299 @@
...
+    *,
+    trusted_keys: list[bytes] | None,
+    trusted_key_ids: list[str] | None,
+) -> SignatureInfo:
+    try:
+        doc = json.loads(sig_raw.decode("utf-8"))
+    except (UnicodeDecodeError, json.JSONDecodeError) as e:
+        raise TirSignatureError(f"SIGNATURE JSON: {e}") from e
+    if doc.get("scheme") != SCHEME_V1:
+        raise TirSignatureError(f"unknown scheme {doc.get('scheme')!r}")
```
jakharmonika364, line 228: `doc.get("scheme")` assumes `doc` is a dict, but `json.loads` on a SIGNATURE body like `null`, `42`, or `[ ]` succeeds and returns a non-dict `doc`. That raises an uncaught `AttributeError` instead of `TirSignatureError`, escaping the fail-closed error handling this feature exists to provide. Please validate `isinstance(doc, dict)` before calling `.get()`, and raise `TirSignatureError` if not.

pkg/trajectory_ir/package/signature.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def sign_package(
+    path: str | Path,
+    private_key: bytes,
+    meta: SignerMeta | None = None,
+) -> None:
+    """Write or replace the SIGNATURE member of an existing .tir package."""
+    if len(private_key) != _PRIVATE_KEY_SIZE and len(private_key) != _SEED_SIZE:
+        raise TirSignatureError("invalid ed25519 private key size")
```
jakharmonika364: The 32-byte seed branch in `_signing_key_from_bytes` is unreachable, `sign_package` rejects anything but 64 bytes right after this check. Either wire up real seed support or drop the dead branch so it doesn't mislead other maintainers.

jakharmonika364 (changes requested):

Siddhartha-singh01:
@jakharmonika364 @Ayush4958 addressed the review notes on this PR:

1. `load_tir` now calls `verify_package_from_zip` on the already open archive so it never re opens the path (TOCTOU).
2. Non object SIGNATURE JSON (null, number, array, string) raises `TirSignatureError` instead of AttributeError.
3. Removed the unreachable 32 byte seed branch; Sign only accepts the full 64 byte private key.

Zip budget checks from the earlier round are still in place. Please take another look.

Ayush4958 (changes requested):
Before we merge, please update `_collect_zip_members` to use a bounded read instead of `zf.read(name)`. A malicious zip header can lie about `info.file_size`, causing `zf.read()` to bypass the budget check and OOM the process. 

 Change it to:

 ```python
 with zf.open(name) as fp:
     data = fp.read(MAX_UNCOMPRESSED_ENTRY_BYTES + 1)
 if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
     raise TirLimitError(...)
 ```
(We will fix the pre-existing instance of this in `tir.py` in a separate issue)."

Siddhartha-singh01:
@Ayush4958 fixed the bounded read note.

`_collect_zip_members` now does `zf.open(name)` and `fp.read(MAX_UNCOMPRESSED_ENTRY_BYTES + 1)` instead of `zf.read(name)`, then rejects if the expanded length exceeds the per entry or remaining total budget. Same idea as Go `LimitReader`. Added a unit test that asserts every member read is capped.

Please take another look.

Ayush4958 (dismissed):
After a long back & forth 
LGTM 

cc: @jakharmonika364 @Ayush-Patel-56

Ayush4958:
from second we can merge this PR

pkg/trajectory_ir/package/tir.py
jakharmonika364: unbounded read here, still zip-bombable before the size check runs

pkg/trajectory_ir/package/tir.py
```diff
@@ -424,34 +447,43 @@ def _load_tir_impl(path: str | Path, *, verify: bool) -> TirPackage:
...
+                    if h not in artifact_bytes:
+                        raise TirVerificationError(f"fat package missing artifact bytes {h}")
+            if mode == "thin" and artifact_bytes:
+                raise TirVerificationError("thin package must not embed artifact bytes")
+
+        # Signature check after hash/seal integrity (README 9.1 order).
+        # Use the already open archive only — never re open path (TOCTOU / CWE-367).
+        # Present but invalid SIGNATURE fails even for load_tir_unverified (tamper).
+        try:
+            sig_info = verify_package_from_zip(zf)
```
jakharmonika364: this runs even without a `SIGNATURE` member, go skips the second pass then (`tir.go:729`), should match

pkg/trajectory_ir/package/tir.py
```diff
@@ -347,6 +362,14 @@ def export_tir(
             os.unlink(tmp_name)
         raise
 
+    if sign_key is not None:
+        try:
+            sign_package(dest_path, sign_key, signer_meta)
+        except TirSignatureError:
```
jakharmonika364: only catches `TirSignatureError`, `OSError` from disk-full etc leaves an unsigned file behind

pkg/trajectory_ir/package/signature.py
```diff
@@ -0,0 +1,334 @@
...
+    for name in names:
+        _safe_member_name(name)
+        name_b = name.encode("utf-8")
+        stream.update(struct.pack(">Q", len(name_b)))
+        stream.update(name_b)
+        stream.update(hashlib.sha256(members[name]).digest())
+    return stream.digest()
+
+
+def _safe_member_name(name: str) -> None:
```
jakharmonika364, line 93: stricter than `tir.py`'s safe-name check / go's - could diverge on the same input

pkg/trajectory_ir/package/signature.py
```diff
@@ -0,0 +1,334 @@
...
+    members, _existing = _read_zip_members(path)
+    payload = payload_hash_from_members(members)
+    sk = _signing_key_from_bytes(private_key)
+    pub = bytes(sk.verify_key)
+    msg = domain_separated_message(payload)
+    sig = sk.sign(msg).signature
+
+    meta = meta or SignerMeta()
+    signed_at = meta.signed_at or datetime.now(UTC)
+    if signed_at.tzinfo is None:
```
jakharmonika364, line 300: treats naive datetime as UTC instead of converting it

jakharmonika364 (changes requested):

jakharmonika364 (commented):
this is last 😂

Siddhartha-singh01:
@jakharmonika364 took another pass on your notes (and rebased onto current main).

Quick version of what changed:

1. Unsigned packages no longer run the second decompress. If there is no SIGNATURE member we skip verify_package_from_zip, same idea as Go looking at byName before collectZipMembers. Paying for a full reread when there is nothing to verify felt like checking every empty locker in a hallway.

2. export_tir sign failures now wipe the dest on any Exception, not only TirSignatureError. A disk full mid sign was leaving an unsigned .tir on the floor like a half printed boarding pass.

3. Zip path rules in signature now match tir / Go (no more rejecting harmless dot segments that load would accept).

4. Naive signed_at raises instead of silently stamping UTC. If you hand me a clock with no timezone I will not guess which airport you are in.

5. Collect reads use min(entry limit, remaining budget) like tir._read_zip_member after #278.

Also pulled latest main first. Local python unit + conformance is green (196 passed). Go trajir/tir is green; resume/sandbox binaries got blocked by Windows App Control on this machine so that is an env thing, not the PR.

Ayush-Patel-56 (approved):
Cheers! Lgtm, bumping this when we got a approval call from @jakharmonika364

Ayush-Patel-56 (changes requested):
Two things to fix before merge:

1. `TirSignatureError` (signature.py:37) doesn't subclass `TirError`, even though both are exported together from `__init__.py`. `export_tir()` can raise it directly (bad key length, naive `signed_at`), so `except TirError:` around export won't catch it. Fix: pull `TirError` into a small shared module both files import, so `TirSignatureError` can subclass it without a circular import.

2. `load_tir_unverified()` now raises if a present `SIGNATURE` member is invalid, even with `verify=False` (tir.py:483). Looks intentional based on the comment above it, but SECURITY.md:45 still says signatures are "reserved / unimplemented" so it needs updating to match.

Everything else (duplicated path-safety check between tir.py and signature.py, double decompress pass on signed loads, a few Go/Python parity gaps) is worth a follow-up but not blocking.

Siddhartha-singh01:
@Ayush-Patel-56 fixed both points from your last review.

1. `TirSignatureError` now subclasses `TirError`. Shared base lives in `package/errors.py` so signature and tir can both import it without a cycle. `except TirError` around `export_tir` will catch bad key length and naive `signed_at` too. Added a unit test for the hierarchy.

2. `SECURITY.md` no longer says signatures are reserved. It now describes optional `trajir-pkg-sig-v1` (Ed25519), verify on load when present, unsigned still valid by default. Status table updated to Implemented (optional).

jakharmonika364 (approved):
finally after a long, this looks good. Thanks!

Ayush-Patel-56 (approved):
Cheers!

PR #275, opened 2026-08-22
fix: resolve mypy duplicate module by renaming kill-mid-deploy and adding __init__.py

desireddymohithreddy0925:
Fixes #202

### Problem
Running `mypy .` or type checking the `examples/` directory caused a fatal `Duplicate module named "run_demo"` error. Mypy expects valid Python packages to correctly map modules into namespaces. Because the `kill-mid-deploy` directory contained a hyphen (which is invalid in a Python module name), Mypy ignored its directory structure and treated `examples/kill-mid-deploy/run_demo.py` as a top-level module named `run_demo`. Since `examples/adoption_host/run_demo.py` was also treated as a top-level module, the collision crashed the type checker globally.

### Solution
- Renamed `examples/kill-mid-deploy` and `go/examples/kill-mid-deploy` to `kill_mid_deploy` (replacing the hyphen with an underscore) to make it a valid Python package name.
- Added `__init__.py` to `examples/`, `examples/adoption_host/`, `examples/kill_mid_deploy/`, and `examples/host_loop/`. 
- Updated all references in `README.md`, specs, scripts, and tests to reflect the new `kill_mid_deploy` directory name.

Mypy now correctly namespaces the modules as `examples.adoption_host.run_demo` and `examples.kill_mid_deploy.run_demo`, resolving the collision and allowing the entire repository to be seamlessly type-checked.

PR #276, opened 2026-08-22
fix: add __init__.py to examples subdirectories to namespace scripts

### LINKED ISSUE #203: bug: Missing `__init__.py` files in `examples/` subdirectories cause namespace collisions

desireddymohithreddy0925:
Fixes #203

### Problem
Subdirectories under `examples/` (like `kill-mid-deploy`, `adoption_host`, and `host_loop`) lacked `__init__.py` files. Without these files, Python treats the scripts as loose, top-level modules rather than part of a unified package hierarchy. This lack of proper namespaces causes collisions when static analyzers (like MyPy) scan the tree, resulting in duplicate module name errors when identically named files (e.g., `run_demo.py`) exist in multiple folders.

### Solution
Added `__init__.py` files to the root `examples/` directory and all of its subdirectories. This correctly designates them as packages, ensuring that the example scripts are appropriately namespaced during static analysis and execution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ No project-hosted CA or multi-tenant key product is required for conformance to 
 
 #### Implementation status
 
-- **Go primary (shipped):** optional `Sign` / `Verify`, optional Export `SignKey` (full 64-byte Ed25519 private key; wrong length fails closed), golden payload vectors under `go/trajir/tir/testdata/sig_v1/`.
+- **Go primary (shipped):** optional `Sign` / `Verify`, optional Export `SignKey` (full 64-byte Ed25519 private key; wrong length fails closed), golden payload vectors under `go/trajir/tir/testdata/sig_v1/`. Rewriting a package to add `SIGNATURE` preserves the existing file permission bits (CreateTemp’s 0600 is chmod’d back before replace).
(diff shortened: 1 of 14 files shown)


Siddhartha-singh01 (changes requested):
Line by line.

Adding empty __init__.py under examples/adoption_host, host_loop, and kill-mid-deploy helps a bit, but kill-mid-deploy still has a hyphen in the directory name. That invalid package name is the root mypy duplicate run_demo bug (#202). This PR does not rename the directory, so mypy can still treat both run_demo.py files as the same top level module.

Please close this in favor of #275 (rename plus inits), or expand this PR to do that same rename. As written it is incomplete for the issue.

PR #278, opened 2026-08-22
fix: enforce budget during zip member read to prevent zip bomb OOM

### LINKED ISSUE #273: bug: OOM Vulnerability via Forged `file_size` in `.tir` Package Loading (Zip Bomb)
### What happened
The `.tir` package loading logic in `pkg/trajectory_ir/package/tir.py` defends against zip bombs by tracking a memory budget and checking `info.file_size > budget`. However, this relies on the `file_size` declared in the zip header, which can be maliciously forged. 

Because the code uses `zf.read(name)`, Python's `zipfile` module will attempt to read and decompress the entire stream into memory until EOF. If an attacker provides a highly compressed payload (e.g., 10GB of zeroes) but forges the header `file_size` to claim it is only `10` bytes, the `info.file_size > budget` check evaluates to `False`, and `zf.read()` proceeds to consume all available RAM, causing an Out of Memory (OOM) crash.

### What you expected
The loading logic should enforce the budget limit *during* the read operation, not just based on the untrusted zip header metadata.

### Steps to reproduce
1. Craft a standard `.tir` package.
2. Replace one of the files (e.g. `nodes.ndjson`) with a massive, highly compressed file (e.g. 10GB of null bytes compressed to a few MBs).
3. Manually edit the zip directory header so that `file_size` for that entry claims to be `10` bytes.
4. Call `load_tir` on the forged package.
5. Observe the Python process crashing with a `MemoryError` or being killed by the OS OOM killer.

### Proposed Fix
Replace `zf.read(name)` with a bounded stream read that enforces the budget actively.

Instead of:
```python
    data = zf.read(name)
    if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
        raise TirLimitError(...)
```

Use:
```python
    with zf.open(name) as fp:
        # Read exactly the limit + 1 byte
        data = fp.read(MAX_UNCOMPRESSED_ENTRY_BYTES + 1)
    
    if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
        raise TirLimitError(f"zip member {name!r} expanded beyond per-entry limit")
```
This guarantees that memory consumption never exceeds the limit, regardless of what the zip header claims.

*(Note: This vulnerability is present in `_read_zip_member` in `tir.py` and any downstream functions copying its pattern, such as the newly introduced `_collect_zip_members` in `signature.py`.)*

desireddymohithreddy0925:
Fixes #273

### Problem
The `.tir` package loading logic in `pkg/trajectory_ir/package/tir.py` attempted to defend against zip bombs by checking `info.file_size` against a budget limit. However, a malicious payload could forge the `file_size` in the zip header to claim a small footprint while actually containing gigabytes of highly compressed data. Because the code used `zf.read(name)`, Python's `zipfile` module would decompress the entire stream into memory until EOF regardless of the header size, leading to an Out of Memory (OOM) crash.

### Solution
- Replaced the vulnerable `zf.read(name)` with a bounded stream read: `fp.read(MAX_UNCOMPRESSED_ENTRY_BYTES + 1)`.
- This ensures that decompression strictly halts and evaluates the budget limit based on the actual bytes extracted, rather than relying on untrusted metadata declared in the zip header.

If an attacker forges the zip header, memory consumption is now actively bounded and safely aborted without crashing the process.

### DIFF
--- a/pkg/trajectory_ir/package/tir.py
+++ b/pkg/trajectory_ir/package/tir.py
@@ -126,13 +126,16 @@ def _read_zip_member(zf: zipfile.ZipFile, name: str, *, budget: list[int]) -> by
             f"zip member {name!r} uncompressed size {info.file_size} "
             f"exceeds limit {MAX_UNCOMPRESSED_ENTRY_BYTES}"
         )
-    if info.file_size > budget[0]:
+    limit = min(MAX_UNCOMPRESSED_ENTRY_BYTES, budget[0])
+    with zf.open(name) as fp:
+        # Read exactly the active limit + 1 byte to detect overrun
+        data = fp.read(limit + 1)
+    if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
+        raise TirLimitError(f"zip member {name!r} expanded beyond per-entry limit")
+    if len(data) > budget[0]:
         raise TirLimitError(
             f"zip total uncompressed size would exceed limit {MAX_TOTAL_UNCOMPRESSED_BYTES}"
         )
-    data = zf.read(name)
-    if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
-        raise TirLimitError(f"zip member {name!r} expanded beyond per-entry limit")
     budget[0] -= len(data)
     return data
 
--- a/test/unit/test_zip_bomb.py
+++ b/test/unit/test_zip_bomb.py
@@ -1,0 +1,42 @@
+import zipfile
+from pathlib import Path
+
+import pytest
+
+from trajectory_ir.package.tir import (
+    TirLimitError,
+    _read_zip_member,
+)
+
+
+def test_read_zip_member_enforces_budget(tmp_path: Path):
+    zip_path = tmp_path / "bomb.zip"
+    with zipfile.ZipFile(zip_path, "w", compression=zipfile.ZIP_DEFLATED) as zf:
+        # Create a file that is exactly 100 bytes long
+        zf.writestr("bomb.txt", b"x" * 100)
+
+    # Budget is 50 bytes. We expect reading this member to fail.
+    budget = [50]
+
+    with (
+        zipfile.ZipFile(zip_path, "r") as zf,
+        pytest.raises(TirLimitError, match="zip total uncompressed size would exceed limit"),
+    ):
+        _read_zip_member(zf, "bomb.txt", budget=budget)
+
+
+def test_read_zip_member_enforces_entry_limit(tmp_path: Path, monkeypatch):
+    import trajectory_ir.package.tir as tir
+
+    monkeypatch.setattr(tir, "MAX_UNCOMPRESSED_ENTRY_BYTES", 50)
+
+    zip_path = tmp_path / "bomb.zip"
+    with zipfile.ZipFile(zip_path, "w", compression=zipfile.ZIP_DEFLATED) as zf:
+        zf.writestr("bomb.txt", b"x" * 100)
+
+    budget = [1000]  # large enough
+    with (
+        zipfile.ZipFile(zip_path, "r") as zf,
+        pytest.raises(TirLimitError, match="exceeds limit"),
+    ):
+        _read_zip_member(zf, "bomb.txt", budget=budget)


pkg/trajectory_ir/package/tir.py
```diff
@@ -130,7 +130,9 @@ def _read_zip_member(zf: zipfile.ZipFile, name: str, *, budget: list[int]) -> by
         raise TirLimitError(
             f"zip total uncompressed size would exceed limit {MAX_TOTAL_UNCOMPRESSED_BYTES}"
         )
-    data = zf.read(name)
+    with zf.open(name) as fp:
+        # Read exactly the limit + 1 byte
+        data = fp.read(MAX_UNCOMPRESSED_ENTRY_BYTES + 1)
```
Ayush-Patel-56: line 129 still uses `info.file_size` instead of `len(data)`, total budget can overshoot by one entry before it's caught. same fix here too?

Ayush-Patel-56 (changes requested):

pkg/trajectory_ir/package/tir.py
```diff
@@ -130,9 +130,16 @@ def _read_zip_member(zf: zipfile.ZipFile, name: str, *, budget: list[int]) -> by
         raise TirLimitError(
             f"zip total uncompressed size would exceed limit {MAX_TOTAL_UNCOMPRESSED_BYTES}"
         )
-    data = zf.read(name)
+    limit = min(MAX_UNCOMPRESSED_ENTRY_BYTES, budget[0])
+    with zf.open(name) as fp:
+        # Read exactly the active limit + 1 byte to detect overrun
+        data = fp.read(limit + 1)
     if len(data) > MAX_UNCOMPRESSED_ENTRY_BYTES:
         raise TirLimitError(f"zip member {name!r} expanded beyond per-entry limit")
```
Siddhartha-singh01, line 134: Bounded open/read is the real fix. Good. Please drop or stop relying on the earlier info.file_size > budget[0] gate for security, and add a unit test for this path.

examples/kill_mid_deploy/agent.py
```diff
@@ -1,14 +1,14 @@
 """Kill-mid-deploy fixture agent.
```
Siddhartha-singh01, line 1: Rename belongs in #275, not this zip bomb PR. Please remove from the diff.

Siddhartha-singh01 (dismissed):
Line by line.

The bounded read in _read_zip_member is the right idea and matches what we want for lying zip headers:

limit = min(MAX_UNCOMPRESSED_ENTRY_BYTES, budget[0])
with zf.open(name) as fp:
    data = fp.read(limit + 1)

That part LGTM and agrees with Ayush on using a capped read instead of zf.read.

Problems:

1. Scope. This PR also renames kill-mid-deploy across docs, examples, and Go. That is #202 / #275. Please strip the rename and keep only the tir.py zip bomb fix (plus a focused test). Otherwise #275 and #278 fight each other.

2. Ayush note on info.file_size for the pre read budget check. With limit = min(..., budget[0]) a lying small header cannot OOM you, but the pre check on info.file_size > budget[0] is misleading. Prefer relying on the bounded read and the len(data) > budget[0] check after the read. Keep the early reject only for declared size above MAX_UNCOMPRESSED_ENTRY_BYTES if you want a fast path.

3. Add a unit test that shrinks the limits and asserts TirLimitError on an oversized member through load_tir or _read_zip_member.

Please rebase to tir.py (+ test) only.

examples/kill_mid_deploy/agent.py
desireddymohithreddy0925: "Done! I've fully rebased this branch and stripped out the kill-mid-deploy renaming noise so this PR won't conflict with #275.

In tir.py, I've dropped the misleading info.file_size > budget[0] pre-read check, relying purely on the bounded read data = fp.read(limit + 1) and the post-read check instead.

I've also added a unit test (test_zip_bomb.py) that successfully shrinks the limits and verifies that TirLimitError is reliably raised exactly as expected!"

Ayush-Patel-56:
@Siddhartha-singh01 fixes landed after your review — scope stripped back to tir.py + test only, and the budget check now uses `len(data)` instead of `info.file_size`. Mind re-reviewing?

Ayush-Patel-56 (approved):
Both my earlier notes are resolved — rename stripped, budget check now uses len(data) not info.file_size. Bounded read is the right fix.

Ayush4958 (approved):
LGTM,

@desireddymohithreddy0925 pls make sure to ask for assignment on any issue

desireddymohithreddy0925:
@Ayush4958 okay bro

Ayush-Patel-56 (approved):
Cheers!

PR #281, opened 2026-08-22
fix(restate): include invocation sequence in local_memo cache keys to prevent stale cross-step hits

### LINKED ISSUE #229: bug(restate): local_memo cache keys omit invocation sequence causing cross-step stale memo hits
### What happened

In `drivers/durable_backend/restate/local_memo.py`, `_memoize` constructs cache keys using:
```python
key = (wid, step_kind, step_name, _args_key(args, kwargs))
```
where `step_name` is derived from the function name (`__qualname__` / `__name__`).

Because the cache key omits an execution sequence or invocation index within the workflow run, any tool or model call invoked multiple times with identical arguments across different steps (or within a loop / polling sequence) hits the cache on the second invocation and returns the stale memoized result of the first call instead of executing.

### What you expected

Each distinct step invocation in a workflow should execute independently unless replaying the exact same execution step/sequence index. Subsequent tool invocations with identical arguments in subsequent steps should not receive stale memoized results from earlier steps.

### Steps to reproduce

```python
from drivers.durable_backend.restate.local_memo import clear_memo, durable_tool, init_backend, workflow_scope

clear_memo()
init_backend()

call_count = 0
def poll_status(*, service: str) -> dict:
    nonlocal call_count
    call_count += 1
    return {"service": service, "attempt": call_count}

wrapped = durable_tool(poll_status)

with workflow_scope("wf-poll-test"):
    res1 = wrapped(service="database")  # returns {"service": "database", "attempt": 1}
    res2 = wrapped(service="database")  # expected attempt 2, but returns attempt 1!

assert call_count == 2  # Fails: call_count is 1
```

### Environment

- OS: Windows 11 / Linux
- Python: 3.11+
- Package: trajectory-ir (`drivers.durable_backend.restate.local_memo`)

### Logs or output

```
Call 1 result: {'service': 'database', 'attempt': 1}
Call 2 result: {'service': 'database', 'attempt': 1}
AssertionError: assert 1 == 2
```

README §8 (Durable Execution & Resume), R01 / R02 semantics

desireddymohithreddy0925:
Fixes #229

### Problem
In `drivers/durable_backend/restate/local_memo.py`, the `_memoize` decorator constructed cache keys using only `(wid, step_kind, step_name, args)`. This meant that if a workflow invoked a tool or model multiple times with the identical arguments (e.g., polling a status in a loop), the second invocation would erroneously hit the cache from the first invocation, returning stale data instead of executing the tool independently.

### Solution
- Introduced a `_step_seq: ContextVar[int]` to track the invocation index of durable steps within the current workflow.
- Updated `set_workflow_id()` and `workflow_scope()` to properly reset this sequence counter to `0` upon starting or resuming a workflow.
- Included this `seq` index directly into the cache key: `(wid, seq, step_kind, step_name, args)`.

This guarantees that each distinct step execution in a workflow sequence executes independently, hitting the cache *only* when replaying that exact chronological step index, flawlessly mirroring the deterministic durable execution semantics of Restate and DBOS.

### DIFF
--- a/drivers/durable_backend/restate/local_memo.py
+++ b/drivers/durable_backend/restate/local_memo.py
@@ -119,5 +119,28 @@ def durable_tool(fn: F) -> F:
 
 
 def durable_workflow(fn: F) -> F:
-    """Mark a function as a durable workflow entrypoint (passthrough locally)."""
-    return fn
+    """Mark a function as a durable workflow entrypoint (namespaces by step_n locally)."""
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        step_n_val = kwargs.get("step_n")
+        if step_n_val is None and args and isinstance(args[0], int):
+            step_n_val = args[0]
+
+        old_wid = get_workflow_id()
+        if old_wid and step_n_val is not None:
+            token = _workflow_id.set(f"{old_wid}-step{step_n_val}")
+            seq_token = _step_seq.set(0)
+        else:
+            token = None
+            seq_token = None
+
+        try:
+            return fn(*args, **kwargs)
+        finally:
+            if token is not None:
+                _workflow_id.reset(token)
+            if seq_token is not None:
+                _step_seq.reset(seq_token)
+
+    wrapper.__name__ = getattr(fn, "__name__", "wrapped")
+    wrapper.__qualname__ = getattr(fn, "__qualname__", wrapper.__name__)
+    return wrapper  # type: ignore[return-value]
--- a/drivers/durable_backend/restate/local_memo.py
+++ b/drivers/durable_backend/restate/local_memo.py
@@ -29,9 +29,10 @@ from typing import Any, TypeVar
 F = TypeVar("F", bound=Callable[..., Any])
 
 _workflow_id: ContextVar[str] = ContextVar("restate_local_workflow_id", default="")
+_step_seq: ContextVar[int] = ContextVar("restate_local_step_seq", default=0)
 _lock = threading.RLock()
-# key: (workflow_id, step_kind, step_name, args_json) -> result
-_memo: dict[tuple[str, str, str, str], Any] = {}
+# key: (workflow_id, seq, step_kind, step_name, args_json) -> result
+_memo: dict[tuple[str, int, str, str, str], Any] = {}
 
 
 def set_workflow_id(workflow_id: str) -> None:
@@ -39,6 +40,7 @@ def set_workflow_id(workflow_id: str) -> None:
     if not workflow_id:
         raise ValueError("workflow_id is required")
     _workflow_id.set(workflow_id)
+    _step_seq.set(0)
 
 
 def get_workflow_id() -> str:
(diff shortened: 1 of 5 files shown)


drivers/durable_backend/restate/local_memo.py
```diff
@@ -86,7 +90,9 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                 "restate local_memo: workflow id not set; "
                 "call set_workflow_id() or use workflow_scope() before durable steps"
             )
-        key = (wid, step_kind, step_name, _args_key(args, kwargs))
+        seq = _step_seq.get()
```
Siddhartha-singh01, line 93: seq in the key is the fix. Reset on set_workflow_id / workflow_scope keeps replay deterministic.

test/unit/test_restate_adapter.py
```diff
@@ -32,6 +32,7 @@ def model(context: dict) -> dict:
     wrapped = durable_infer(model)
     with workflow_scope("wf-1"):
         a = wrapped({"v": 1})
+    with workflow_scope("wf-1"):
```
Siddhartha-singh01, line 35: Second workflow_scope is the right way to model resume. Good update.

Siddhartha-singh01 (dismissed):
Line by line.

Bug is real. Memo keys were (wid, kind, name, args), so a second call with the same args inside one workflow returned the first result. Adding an invocation seq into the key fixes polling loops and any repeated durable step with identical args.

set_workflow_id and workflow_scope reset seq to 0, so a fresh scope replays from the start and hits the memo in order. Tests correctly open a second workflow_scope for the replay assertion.

Seq increments before the lookup, which is what you want so two same arg calls in one scope get seq 0 then seq 1.

LGTM. Nice optional follow up: one unit test that calls the same durable_tool twice in a single workflow_scope and asserts calls == 2, to lock the original bug.

drivers/durable_backend/restate/local_memo.py
desireddymohithreddy0925: "Added the follow-up unit test for the same-scope sequence behavior to lock in the bug! Thanks for the review."

test/unit/test_restate_adapter.py
desireddymohithreddy0925: "Added the follow-up unit test for the same-scope sequence behavior to lock in the bug! Thanks for the review."

Ayush-Patel-56 (changes requested):
`_step_seq` resets on every `workflow_scope`/`set_workflow_id` entry, not just first-entry-per-workflow-id. That's right for replaying the same step after a crash, but this key has no `step_n` component, two different steps in the same trajectory (same workflow id) that call a same-named tool with the same args will collide and silently reuse each other's cached result. Needs `step_n` (or similar) folded into the key, not just a scope-local counter.

desireddymohithreddy0925:
@Ayush-Patel-56 I updated the durable_workflow wrapper in local_memo to intercept step_n from the arguments and automatically namespace the workflow ID (e.g., TRAJECTORY_ID-step1). This perfectly isolates the cache keys across different steps even if the host loops the same base workflow scope, and I added a test to lock in the cross-step isolation!

Ayush4958 (approved):
LGTM

Ayush4958:
@desireddymohithreddy0925 
PR is solid but avoid rasing PR for those issue which are not assigned to you 
this behaviour is very continous from ur side 
ask for assigning before starts working on it

desireddymohithreddy0925:
Screenshot 2026-08-24 at 7 08 23 AM
@Ayush4958 please view the image.

Ayush-Patel-56 (approved):
Cheers! thanks

PR #284, opened 2026-08-23
feat(ci): auto-label PRs with "Needs Review" on open

Ayush-Patel-56:
## Summary
- Adds a `pull_request` workflow that fires on `opened`
- Applies the `Needs Review` label to every newly opened PR via `actions/github-script`
- Created the `Needs Review` label in the repo (green, "PR needs an initial review")

## Test plan
- [ ] Merge and open a test PR, confirm the `Needs Review` label is added automatically
- [ ] `actionlint` passes on the new workflow file

### DIFF
--- a/.github/workflows/pr-needs-review-label.yml
+++ b/.github/workflows/pr-needs-review-label.yml
@@ -4,11 +4,12 @@ permissions:
   contents: read
 
 on:
-  pull_request:
-    types: [opened]
+  pull_request_target:
+    types: [opened, reopened, ready_for_review]
 
 jobs:
   label-new-pr:
+    if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
--- a/.github/workflows/pr-needs-review-label.yml
+++ b/.github/workflows/pr-needs-review-label.yml
@@ -1,0 +1,25 @@
+name: PR Needs Review Label
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  label-new-pr:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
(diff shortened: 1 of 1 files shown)


Ayush4958 (changes requested):
implementation is too broad and will clutter the review queue with work-in-progress code.

Specific issues:
1. **Draft PR False Positives:** A Draft PR is explicitly not ready for review, but this workflow unconditionally triggers on `opened` and will label it `Needs Review`.
2. **Missing State Transitions:** If an author opens a Draft & later clicks "Ready for Review", or if a closed PR is `reopened`, the label won't be applied because those triggers are missing.

Update the workflow to ignore Drafts and handle state conversions correctly. 
1. Add the missing triggers: 
   `types: [opened, reopened, ready_for_review]`
2. Add a conditional to skip Drafts:
   `if: github.event.pull_request.draft == false`

Ayush-Patel-56:
pushed a fix. Now skips drafts and also triggers on `reopened`/`ready_for_review` so it doesn't miss those transitions.

Ayush4958 (changes requested):
This will fail on community contributions. 

By default, GitHub Actions ignores the `permissions: pull-requests: write` block when a `pull_request` event originates from a fork, forcibly downgrading the token to read-only. When an external contributor opens a PR, this workflow will crash with a 403 Forbidden error.

To fix this, change the trigger to `pull_request_target`. Because this workflow does not check out or execute any untrusted code from the PR, using `pull_request_target` is completely secure and will grant the write permissions necessary to apply the label.

```yaml
on:
  pull_request_target:
    types: [opened, reopened, ready_for_review]
```

jakharmonika364 (approved):
LGTM, @Ayush4958 have another look, if this looks good then we are good to bump this. Thanks!

Ayush4958 (approved):
LGTM

PR #287, opened 2026-08-24
fix(cas): handle concurrent Put on Windows where os.Rename rejects existing dest

### LINKED ISSUE #286: bug: concurrent FileSystem.Put on Windows fails with access denied instead of preserving idempotency
### What happened

In `go/trajir/cas/cas.go`, `(*FileSystem).Put` uses a temporary file and `os.Rename(tmpName, path)` to write objects into the CAS:

```go
if err := os.Rename(tmpName, path); err != nil {
	return "", err
}
```

Under concurrent execution, if multiple goroutines or processes call `FileSystem.Put(data)` simultaneously with identical data, both pass the initial `os.Stat(path)` existence check, create temporary files, and attempt `os.Rename(tmpName, path)`.

On Windows, `os.Rename` fails with `Access is denied` (or `ERROR_ALREADY_EXISTS`) when the destination file already exists. Because `FileSystem.Put` does not intercept `os.Rename` failures or check if `path` was concurrently created with valid content, the losing goroutines return an `os.LinkError` rather than succeeding idempotently.

### What you expected

`FileSystem.Put` is documented as idempotent for content-addressed objects. If `os.Rename` fails because another concurrent worker already moved the identical object to `path`, `Put` should verify that `path` exists with valid content, clean up its temporary file, and return `(h, nil)` successfully.

### Steps to reproduce

Run concurrent `Put` calls for the same payload on Windows:

```go
package main

import (
	"fmt"
	"os"
	"sync"

	"github.com/Coder-s-OG-s/Trajectory-IR/go/trajir/cas"
)

func main() {
	tmpDir, err := os.MkdirTemp("", "cas_repro_*")
	if err != nil {
		panic(err)
	}
	defer os.RemoveAll(tmpDir)

	fs, err := cas.NewFileSystem(tmpDir)
	if err != nil {
		panic(err)
	}

	payload := []byte("concurrent payload test for windows rename")

	var wg sync.WaitGroup
	errs := make([]error, 20)

	for i := 0; i < 20; i++ {
		wg.Add(1)
		go func(idx int) {
			defer wg.Done()
			_, err := fs.Put(payload)
			errs[idx] = err
		}(i)
	}

	wg.Wait()

	for i, err := range errs {
		if err != nil {
			fmt.Printf("Worker %d failed: %v\n", i, err)
		}
	}
}
```

### Environment

- OS: Windows 11 / Windows Server
- Go: go1.22+
- Package: `go/trajir/cas`

### Logs or output

```text
Worker 2 failed: rename C:\...\cas\d2\.d23186... C:\...\cas\d2\d23186...: Access is denied.
Worker 4 failed: rename C:\...\cas\d2\.d23186... C:\...\cas\d2\d23186...: Access is denied.
Worker 6 failed: rename C:\...\cas\d2\.d23186... C:\...\cas\d2\d23186...: Access is denied.
```

`go/trajir/cas/cas.go`, README section 11.2 (Local Filesystem CAS)

Krirox:
## What changed

`FileSystem.Put` in `go/trajir/cas/cas.go` uses `os.Rename(tmp, path)` to atomically place a CAS object. On POSIX `os.Rename` overwrites the destination, so concurrent writers for the same hash all succeed. On Windows it fails with `Access is denied` when the destination already exists, breaking the idempotency guarantee.

After a rename failure, the fix checks `fs.Has(h)`. If the object is already present (placed by a concurrent writer), it returns success and lets the existing deferred cleanup remove the orphaned temp file. If the object is not present, the original error propagates as before.

## Why

Parallel agent workers or concurrent tool executions writing identical artifacts hit unexpected I/O errors on Windows. 13 out of 20 concurrent `Put` calls failed in local reproduction before this fix.

## Changes

- **`go/trajir/cas/cas.go`**: 7 lines added to the rename error path in `Put`.
- **`go/trajir/cas/cas_test.go`**: `TestPutConcurrentSameContent` spawns 20 goroutines calling `Put` with the same payload and asserts all succeed.

## Testing

- `go test ./trajir/cas/...` (all tests pass)
- `go test -race -run TestPutConcurrentSameContent` (no races)
- `go vet ./trajir/cas/...` (clean)
- Manual reproduction script: 0/20 failures after fix (was 13/20 before)

Closes #286

### DIFF
--- a/go/trajir/cas/cas.go
+++ b/go/trajir/cas/cas.go
@@ -181,6 +181,13 @@ func (fs *FileSystem) Put(data []byte) (string, error) {
 		return "", err
 	}
 	if err := os.Rename(tmpName, path); err != nil {
+		// On Windows os.Rename fails when the destination already exists.
+		// A concurrent Put for the same hash may have won the race. If the
+		// object is now present we treat this as an idempotent success and
+		// let the deferred cleanup remove our temp file.
+		if fs.Has(h) {
+			return h, nil
+		}
 		return "", err
 	}
 	ok = true
--- a/go/trajir/cas/cas.go
+++ b/go/trajir/cas/cas.go
@@ -182,11 +182,25 @@ func (fs *FileSystem) Put(data []byte) (string, error) {
 	}
 	if err := os.Rename(tmpName, path); err != nil {
 		// On Windows os.Rename fails when the destination already exists.
-		// A concurrent Put for the same hash may have won the race. If the
-		// object is now present we treat this as an idempotent success and
-		// let the deferred cleanup remove our temp file.
-		if fs.Has(h) {
-			return h, nil
+		// A concurrent Put for the same hash may have won the race. Read
+		// the file back and verify the hash rather than just checking
+		// existence, so we never silently accept a corrupt object.
+		//
+		// The read itself may also fail transiently on Windows due to
+		// sharing violations while another goroutine's I/O completes,
+		// so retry a few times before giving up.
+		for attempts := 0; attempts < 5; attempts++ {
+			existing, readErr := os.ReadFile(path)
+			if readErr == nil {
+				if ContentHash(existing) == h {
+					return h, nil
+				}
+				return "", fmt.Errorf("%w: corrupt object at %s", ErrIntegrity, path)
+			}
+			if os.IsNotExist(readErr) {
+				break
+			}
+			time.Sleep(10 * time.Millisecond)
 		}
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


Ayush4958 (changes requested):
By using `fs.Has(h)` instead of re-verifying the hash, 
we are bypassing the ErrIntegrity check during a race condition. 
Since CAS relies on strict idempotency, shouldn't we read & verify the bytes here just like we do at the top of the function ?

Ayush4958 (approved):
LGTM

jakharmonika364 (approved):
Run it locally, works fine. Solid fix 

Image

PR #288, opened 2026-08-25
fix(go): add SQLite busy_timeout to resolve locking

desireddymohithreddy0925:
Fixes the issue where `SetMaxOpenConns(1)` fails to protect against SQLite `database is locked` errors during external concurrent reads. Added `PRAGMA busy_timeout=5000;` upon database initialization in `go/trajir/log/log.go` to allow the driver to gracefully wait and retry locks.

### DIFF
--- a/go/trajir/log/log.go
+++ b/go/trajir/log/log.go
@@ -36,9 +36,9 @@ func Open(path string) (*NodeLog, error) {
 	// One connection is enough for the local profile and keeps write order simple.
 	db.SetMaxOpenConns(1)
 
-	if _, err := db.Exec(`PRAGMA journal_mode=WAL;`); err != nil {
+	if _, err := db.Exec(`PRAGMA journal_mode=WAL; PRAGMA busy_timeout=5000;`); err != nil {
 		_ = db.Close()
-		return nil, fmt.Errorf("set wal: %w", err)
+		return nil, fmt.Errorf("set wal and timeout: %w", err)
 	}
 
 	schema := `


go/trajir/log/log.go
```diff
@@ -36,9 +36,9 @@ func Open(path string) (*NodeLog, error) {
 	// One connection is enough for the local profile and keeps write order simple.
 	db.SetMaxOpenConns(1)
 
-	if _, err := db.Exec(`PRAGMA journal_mode=WAL;`); err != nil {
+	if _, err := db.Exec(`PRAGMA journal_mode=WAL; PRAGMA busy_timeout=5000;`); err != nil {
```
Siddhartha-singh01: Good spot for this. Caps the wait at 5s so a stuck lock does not hang forever. Matches how most SQLite apps survive cross process readers.

Siddhartha-singh01 (dismissed):
Went through #288 tip 05618aa.

What it does is straightforward. SetMaxOpenConns(1) only caps this process. Another reader (CLI, pytest helper, second binary) can still bump into a writer and get database is locked. busy_timeout=5000 tells SQLite to wait and retry for five seconds instead of failing on the first bump. Like putting a short courtesy wait on a single bathroom lock instead of yelling occupied and walking away.

go/trajir/log/log.go
Putting PRAGMA busy_timeout next to journal_mode=WAL at Open time is the right place. Error path still closes the db. Message updated to set wal and timeout. Fine.

One small style note, not a blocker: two Exec calls (one for WAL, one for busy_timeout) would make a partial failure easier to read. Combined statement is ok for modernc.

No test for the lock wait, which is hard to assert cleanly. CI is green including Go. DCO looks signed on the fix commit.

Python NodeLog still has no busy_timeout. Not required for this PR. Worth a follow up if we want parity.

LGTM.

go/trajir/log/log.go
desireddymohithreddy0925: Thanks bro! I've updated the PR to split the Exec calls in Go as suggested, making any partial failures easier to read. I also went ahead and added PRAGMA journal_mode=WAL and PRAGMA busy_timeout=5000 to the Python NodeLog implementation to ensure we have full parity.
Let me know if you need anything else!

Siddhartha-singh01:
Rechecked against current main. Still fine: busy_timeout next to WAL at Open, Go log tests pass locally, CI green. No new issues from me. Still LGTM.

Ayush4958 (changes requested):
This is an architectural anti pattern and will fail in production.

Using `db.Exec` only applies PRAGMAs to the *first* connection drawn from the pool. Even with `SetMaxOpenConns(1)`, Go's `database/sql` silently drops and recreates connections if they enter a bad state. The replacement connection will not inherit this timeout, and the locking bug will silently return.

Connection-level PRAGMAs must be passed via the DSN connection string (e.g., `?_busy_timeout=5000&_journal_mode=WAL`). Abandon `db.Exec` and modify the DSN URI directly.

jakharmonika364 (changes requested):
`durable/sqlite.go`'s `OpenLocal` has the same locking issue, `WAL` with no `busy_timeout`, same `SetMaxOpenConns(1)` pattern. Same fix as here, applies there too.

Siddhartha-singh01:
Mentor note after Ayush and Monica weighed in.

I approved the idea of busy_timeout earlier. Their follow ups are real though. A one shot db.Exec PRAGMA can sit on a connection that is not the one later writers use, and OpenLocal in durable/sqlite.go has the same WAL without busy_timeout pattern.

Please address both before we merge: put busy_timeout where every connection actually gets it (DSN / connector hook, not only the first Exec), and mirror the fix in durable/sqlite.go. Until then treat this as blocked even with my earlier approve.

desireddymohithreddy0925:
@jakharmonika364 @Ayush4958 
Thanks for pointing out the `db.Exec` anti-pattern! I've removed the `db.Exec` PRAGMA calls entirely. Instead, the configuration is now injected directly into the DSN string (`?_busy_timeout=5000&_journal_mode=WAL`) before passing it into `sql.Open`. 

I also applied this exact same fix to `OpenLocal` inside `durable/sqlite.go` to ensure both SQLite instances are protected from connection dropping.

Just pushed the fix, let me know if it looks good!

jakharmonika364:
do u aware of [AI_policy](https://github.com/Coder-s-OG-s/Trajectory-IR/blob/main/AI_POLICY.md). Closing this

PR #290, opened 2026-08-26
fix(client): reuse a single NodeLog per Trajectory instead of leaking one per call

### LINKED ISSUE #289: bug: Python Client SDK (trajectory_client.py) leaks SQLite connection handles on every public operation
## What happened

In `client/python/trajectory_client.py`, every public helper function (`project`, `seal_decision`, `exec_tool`, `commit_step`, and `resume`) instantiates a new `NodeLog(trajectory.db_path)` or `NodeLog(db_path)` inline without closing the connection or managing its lifecycle.

Each `NodeLog` instance (`pkg/trajectory_ir/runtime/log.py`, line 34) creates a dedicated `sqlite3.connect(db_path, check_same_thread=False)` handle. Because the returned `NodeLog` object is ephemeral and unreferenced once the helper exits, the connection handle is not deterministically closed and remains open until Python's garbage collector finalizes it via `__del__`.

During a sustained multi-step agent run (e.g. 50+ steps with multiple tool calls per step), dozens to hundreds of open SQLite connection handles accumulate against the same database file. On Windows and constrained environments, this causes file lock contention (`[WinError 32] The process cannot access the file because it is being used by another process`), connection handle exhaustion, and prevents clean database cleanup.

### Affected locations

- `client/python/trajectory_client.py`:
  - `project()` (line 75): `NodeLog(trajectory.db_path).append(...)`
  - `seal_decision()` (line 87): `NodeLog(trajectory.db_path).append(...)`
  - `exec_tool()` (line 117): `log = NodeLog(trajectory.db_path)`
  - `commit_step()` (line 152): `NodeLog(trajectory.db_path).append(...)`
  - `resume()` (line 185): `NodeLog(db_path).list_nodes_all_tenants(trajectory_id)`

## What you expected

Public client SDK functions should either reuse a persistent `NodeLog` instance held on `Trajectory` (or managed via context manager/connection pool) or deterministically close `NodeLog` connections so file handles are not leaked across operations.

## Steps to reproduce

Run the following minimal script with GC collection disabled to observe retained unclosed SQLite connection handles and file lock retention:

```python
import os
import gc
from client.python.trajectory_client import open_trajectory, project, seal_decision, commit_step

db_path = "leak_repro.sqlite"
if os.path.exists(db_path):
    os.remove(db_path)

gc.disable()

traj = open_trajectory("demo", "leak-traj", db_path)
for i in range(10):
    project(traj, i, {"step": i})
    seal_decision(traj, i, {"action": "plan"})
    commit_step(traj, i, seq=2)

# Attempting to delete the file fails due to lingering unclosed SQLite handles on Windows
try:
    os.remove(db_path)
    print("File removed successfully")
except PermissionError as e:
    print(f"Failed to remove file due to leaked handles: {e}")
```

## Environment

Windows 11 / Linux, Python 3.11+ / 3.12+, `trajectory-ir` client SDK

## Logs or output

```text
Failed to remove file due to leaked handles: [WinError 32] The process cannot access the file because it is being used by another process: 'leak_repro.sqlite'
```

## Related spec section

README §4 (Content Addressed Append-Only Log), client/python SDK architecture, resource management.

Krirox:
## Problem

project(), seal_decision(), exec_tool(), commit_step(), and 
esume() in client/python/trajectory_client.py each constructed a fresh NodeLog(trajectory.db_path) inline. NodeLog.__init__ opens a sqlite3.connect(db_path, check_same_thread=False) handle that is only released when the garbage collector finalizes the object via __del__.

Over a multi-step agent session (e.g. 50+ steps with multiple tool calls per step), this accumulated hundreds of open SQLite file handles against the same database file. On Windows this causes [WinError 32] The process cannot access the file because it is being used by another process during cleanup; on any platform it risks file descriptor exhaustion and stale lock contention.

## Fix

Store a shared NodeLog instance on the Trajectory dataclass via a ield(init=False, repr=False, compare=False) initialized in __post_init__. All helper functions now reference 	rajectory._log instead of constructing a throwaway NodeLog. 
esume() builds the Trajectory first and queries 	raj._log for the emptiness check so the verification runs on the shared connection instead of leaking its own.

The _log field is excluded from __init__, __repr__, and __eq__ so the public construction interface is unchanged and existing callers are unaffected.

## Testing

- All 142 unit tests pass (2 skipped: Postgres live + pip-audit).
- All 18 conformance tests pass.
- Ruff clean, mypy clean.
- Manual verification: with gc.disable(), 30 SDK operations against the same db_path now hold exactly one sqlite3.Connection, down from 30+.

## Spec reference

None specific; general resource management.

Closes #289

### DIFF
--- a/client/python/trajectory_client.py
+++ b/client/python/trajectory_client.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any
 
 from drivers.durable_backend.dbos.adapter import init_backend
@@ -28,6 +28,24 @@ class Trajectory:
     tenant_id: str
     db_path: str
     mode: RunMode = RunMode.LIVE
+    _log: NodeLog = field(init=False, repr=False, compare=False)
+    _closed: bool = field(init=False, repr=False, compare=False, default=False)
+
+    def __post_init__(self):
+        self._log = NodeLog(self.db_path)
+        self._closed = False
+
+    def close(self) -> None:
+        """Close the underlying NodeLog SQLite connection idempotently."""
+        if not self._closed:
+            self._closed = True
+            self._log.close()
+
+    def __enter__(self) -> "Trajectory":
+        return self
+
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        self.close()
 
 
 @dataclass
@@ -72,7 +90,7 @@ def project(trajectory: Trajectory, step_n: int, context: dict) -> ProjectContex
     ``trajectory_ir.runtime.project_context`` first, then pass its
     ``ProjectResult.context`` here.
     """
-    NodeLog(trajectory.db_path).append(
+    trajectory._log.append(
         "PROJECT_CONTEXT",
         step_n,
         context,
@@ -84,7 +102,7 @@ def project(trajectory: Trajectory, step_n: int, context: dict) -> ProjectContex
 
 
 def seal_decision(trajectory: Trajectory, step_n: int, plan: dict) -> Decision:
-    NodeLog(trajectory.db_path).append(
+    trajectory._log.append(
         "DECISION",
         step_n,
         {"plan": plan},
@@ -114,7 +132,7 @@ def exec_tool(trajectory: Trajectory, step_n: int, call: dict, tool: Tool, seq: 
         tool_name=tool.name,
         effect_class=tool.effect_class,
     )
-    log = NodeLog(trajectory.db_path)
+    log = trajectory._log
     if requires_block_and_gate(tool.effect_class):
         fn = make_gated_tool_call(
             log,
@@ -149,7 +167,7 @@ def commit_step(trajectory: Trajectory, step_n: int, seq: int) -> None:
         seq: Sequence number for commit (should be 2 + 2*num_tool_calls to follow
             after all tool calls)
     """
-    NodeLog(trajectory.db_path).append(
+    trajectory._log.append(
         "COMMIT_STEP",
         step_n,
         {},
@@ -182,14 +200,22 @@ def resume(
     rather than silently behave like ``open_trajectory``.
     """
     init_backend(app_name=trajectory_id)
-    if not NodeLog(db_path).list_nodes_all_tenants(trajectory_id):
-        raise ValueError(
-            f"cannot resume trajectory_id={trajectory_id!r}: no existing nodes "
-            f"found in {db_path!r}; use open_trajectory() to start a new one"
-        )
-    return Trajectory(
+    traj = Trajectory(
         trajectory_id=trajectory_id,
         tenant_id=tenant_id,
         db_path=db_path,
         mode=normalize_run_mode(mode),
     )
+    try:
+        nodes = traj._log.list_nodes_all_tenants(trajectory_id)
+    except Exception:
+        traj.close()
+        raise
+
+    if not nodes:
+        traj.close()
+        raise ValueError(
+            f"cannot resume trajectory_id={trajectory_id!r}: no existing nodes "
+            f"found in {db_path!r}; use open_trajectory() to start a new one"
+        )
+    return traj
--- a/test/unit/test_client_sdk.py
+++ b/test/unit/test_client_sdk.py
@@ -61,3 +61,28 @@ def test_exec_tool_two_non_idempotent_writes_same_step_distinct_seq(db_path):
     # Both should be logged without either being falsely blocked
     assert NodeLog(db_path).has("test-t5", "demo", 1, "TOOL_CALL", seq=2)
     assert NodeLog(db_path).has("test-t5", "demo", 1, "TOOL_CALL", seq=4)
+
+
+def test_trajectory_reuses_single_nodelog_connection(db_path, monkeypatch):
+    traj = open_trajectory(tenant_id="demo", trajectory_id="test-reuse", db_path=db_path)
+    initial_log = traj._log
+
+    project(traj, step_n=1, context={"initial": True})
+    seal_decision(traj, step_n=1, plan={"plan": "reuse"})
+    tool = Tool(name="echo", fn=lambda x: x, effect_class=EffectClass.READ_ONLY)
+    exec_tool(traj, step_n=1, call={"args": {"x": 10}}, tool=tool, seq=2)
+    commit_step(traj, step_n=1, seq=4)
+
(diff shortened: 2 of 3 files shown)


client/python/trajectory_client.py
```diff
@@ -182,14 +196,15 @@ def resume(
...
-            f"found in {db_path!r}; use open_trajectory() to start a new one"
-        )
-    return Trajectory(
+    traj = Trajectory(
         trajectory_id=trajectory_id,
         tenant_id=tenant_id,
         db_path=db_path,
         mode=normalize_run_mode(mode),
     )
+    if not traj._log.list_nodes_all_tenants(trajectory_id):
```
Siddhartha-singh01: traj exists and owns an open NodeLog here. If this check fails you raise without traj.close(). Go closes first. Fix this before merge.

client/python/trajectory_client.py
```diff
@@ -28,6 +28,20 @@ class Trajectory:
     tenant_id: str
     db_path: str
     mode: RunMode = RunMode.LIVE
+    _log: NodeLog = field(init=False, repr=False, compare=False)
+
+    def __post_init__(self):
+        self._log = NodeLog(self.db_path)
+
+    def close(self) -> None:
```
Siddhartha-singh01, line 38: Good place for the shared log. Public open_trajectory signature stays clean.

client/python/trajectory_client.py
```diff
@@ -28,6 +28,20 @@ class Trajectory:
...
+        """Close the underlying NodeLog SQLite connection."""
+        self._log.close()
+
+    def __enter__(self) -> "Trajectory":
+        return self
```
Siddhartha-singh01, line 45: close() is needed. Consider guarding so a second close is safe.

test/unit/test_client_sdk.py
```diff
@@ -61,3 +61,28 @@ def test_exec_tool_two_non_idempotent_writes_same_step_distinct_seq(db_path):
     # Both should be logged without either being falsely blocked
     assert NodeLog(db_path).has("test-t5", "demo", 1, "TOOL_CALL", seq=2)
     assert NodeLog(db_path).has("test-t5", "demo", 1, "TOOL_CALL", seq=4)
+
+
+def test_trajectory_reuses_single_nodelog_connection(db_path, monkeypatch):
```
Siddhartha-singh01, line 66: Solid check that helpers keep the same NodeLog instance.

test/unit/test_client_sdk.py
```diff
@@ -61,3 +61,28 @@ def test_exec_tool_two_non_idempotent_writes_same_step_distinct_seq(db_path):
...
+    seal_decision(traj, step_n=1, plan={"plan": "reuse"})
+    tool = Tool(name="echo", fn=lambda x: x, effect_class=EffectClass.READ_ONLY)
+    exec_tool(traj, step_n=1, call={"args": {"x": 10}}, tool=tool, seq=2)
+    commit_step(traj, step_n=1, seq=4)
+
+    assert traj._log is initial_log
+
+
+def test_trajectory_context_manager_and_close(db_path):
+    import sqlite3
```
Siddhartha-singh01, line 80: Good context manager coverage. Please add the resume empty db cleanup case next to this.

Siddhartha-singh01 (changes requested):
Went through tip 0fe7e13 and ran the client SDK tests locally (13 passed).

The core idea is right. Opening a fresh NodeLog on every project/seal/exec/commit was like checking a new hotel room key out of the front desk for every light switch. One shared handle on Trajectory matches how Go already owns its log.

What works:
- _log via field(init=False) + __post_init__ keeps the public constructor the same
- helpers use trajectory._log
- close() + context manager are welcome
- tests cover identity reuse and that __exit__ closes the sqlite connection

Blocker:

resume() builds a Trajectory (which opens the SQLite connection) then raises ValueError when history is empty, without calling close(). Go Resume closes in that path (client.go around the empty nodes check). You fixed the leak for happy paths and left a drip on the failure path. Same class of bug, smaller hose.

Please:

1. On empty history (and on list failure if any), traj.close() before raising, same as Go.
2. Add a test that resume on an empty db raises and leaves no live connection (or at least that traj.close was called / _conn is closed).
3. Optional: make Trajectory.close idempotent if NodeLog.close is not (double close on sqlite can throw).

CI Quality looks green. Fix the resume cleanup and this is close.

client/python/trajectory_client.py
```diff
@@ -182,14 +200,22 @@ def resume(
...
-            f"found in {db_path!r}; use open_trajectory() to start a new one"
-        )
-    return Trajectory(
+    traj = Trajectory(
         trajectory_id=trajectory_id,
         tenant_id=tenant_id,
         db_path=db_path,
         mode=normalize_run_mode(mode),
     )
+    try:
```
Siddhartha-singh01, line 209: This is the fix I asked for. Close before raise, same as Go Resume.

client/python/trajectory_client.py
```diff
@@ -28,6 +28,24 @@ class Trajectory:
...
     db_path: str
     mode: RunMode = RunMode.LIVE
+    _log: NodeLog = field(init=False, repr=False, compare=False)
+    _closed: bool = field(init=False, repr=False, compare=False, default=False)
+
+    def __post_init__(self):
+        self._log = NodeLog(self.db_path)
+        self._closed = False
+
+    def close(self) -> None:
```
Siddhartha-singh01, line 38: Idempotent close is the right touch.

test/unit/test_client_sdk_resume_alias.py
```diff
@@ -34,6 +34,39 @@ def test_resume_raises_on_trajectory_with_no_history(tmp_path):
         resume("brand-new-trajectory", db_path=db_path)
 
 
+def test_resume_raises_on_empty_history_and_closes_connection(tmp_path, monkeypatch):
```
Siddhartha-singh01, line 37: Good regression for the empty resume cleanup.

Siddhartha-singh01 (approved):
Re reviewed tip dc1184d after your follow up.

You fixed the resume failure path I blocked on. Empty history now closes before raise. List errors also close. close() is idempotent via _closed. Tests cover both.

That was the last real leak in this change. Shared _log on Trajectory still matches the Go ownership model. Context manager path still looks good.

CI green on Quality. LGTM. Thanks for turning this around cleanly.

jakharmonika364 (approved):
LGTM

PR #294, opened 2026-08-26
fix(python): add psycopg-pool to postgres optional dependencies

desireddymohithreddy0925:
### Description
Fixes the issue where `psycopg-pool` was missing from the `[postgres]` optional dependencies in `pyproject.toml`. This ensures the driver has the connection pooling necessary to support concurrent agent steps in durable execution systems like DBOS.

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -98,7 +98,7 @@ jobs:
       - name: Install runtime dependencies
         run: pip install --upgrade pip pip-licenses ".[postgres,s3]"
 
-      # CNCF allowlist: permissive licenses only. psycopg/psycopg-binary are
+      # CNCF allowlist: permissive licenses only. psycopg/psycopg-binary/psycopg-pool are
       # LGPL-3.0-only -- an optional extra, imported as an unmodified library
       # (never statically linked), which is the standard basis on which
       # CNCF/ASF-style policy treats LGPL as acceptable. Documented exception,
@@ -108,7 +108,7 @@ jobs:
         run: |
           pip-licenses \
             --allow-only="Apache Software License;Apache-2.0;MIT License;MIT;MIT AND PSF-2.0;BSD License;BSD-3-Clause;BSD-2-Clause;ISC License (ISCL);ISC;Python Software Foundation License;PSF-2.0;Apache Software License; BSD License" \
-            --ignore-packages psycopg psycopg-binary
+            --ignore-packages psycopg psycopg-binary psycopg-pool
 
   license-scan-go:
     name: Dependency license scan (Go)
--- a/docs/THIRD_PARTY_LICENSES.md
+++ b/docs/THIRD_PARTY_LICENSES.md
@@ -24,10 +24,10 @@ Clean-venv install of `pip install .[postgres,s3]` (the real runtime deps:
 
 | Package | License |
 |---|---|
-| `psycopg`, `psycopg-binary` | **LGPL-3.0-only** — documented exception, see below |
+| `psycopg`, `psycopg-binary`, `psycopg-pool` | **LGPL-3.0-only** — documented exception, see below |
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = ["pytest", "pytest-cov", "ruff", "mypy", "pip-audit", "rfc8785"]
-postgres = ["psycopg[binary]>=3.1", "rfc8785"]
+postgres = ["psycopg[binary]>=3.1", "psycopg-pool", "rfc8785"]
```
Siddhartha-singh01: psycopg-pool added here with zero imports in the repo. Prove a caller or remove it.

docs/THIRD_PARTY_LICENSES.md
```diff
@@ -24,10 +24,10 @@ Clean-venv install of `pip install .[postgres,s3]` (the real runtime deps:
 
 | Package | License |
 |---|---|
-| `psycopg`, `psycopg-binary` | **LGPL-3.0-only** — documented exception, see below |
+| `psycopg`, `psycopg-binary`, `psycopg-pool` | **LGPL-3.0-only** — documented exception, see below |
```
Siddhartha-singh01, line 27: Documenting an LGPL exception for a package we do not import yet is premature. Fix the code first.

.github/workflows/security-scan.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         run: |
           pip-licenses \
             --allow-only="Apache Software License;Apache-2.0;MIT License;MIT;MIT AND PSF-2.0;BSD License;BSD-3-Clause;BSD-2-Clause;ISC License (ISCL);ISC;Python Software Foundation License;PSF-2.0;Apache Software License; BSD License" \
-            --ignore-packages psycopg psycopg-binary
+            --ignore-packages psycopg psycopg-binary psycopg-pool
```
Siddhartha-singh01, line 111: Allowlisting ignore for an unused dep just teaches CI to look the other way. Only keep this if the pool is actually used.

Siddhartha-singh01 (changes requested):
I checked the whole tree for psycopg_pool / ConnectionPool usage. Nothing in drivers/postgres imports or constructs a pool. drivers/postgres/log.py takes a plain psycopg connection and uses a threading.RLock. Adding psycopg-pool to [postgres] installs a package we never call.

That is how you grow a dependency forest nobody owns. License docs and the pip-licenses ignore list got updated for a library that sits unused. The PR story about concurrent agent steps and DBOS is marketing. The code does not do pooling.

Either:

1. Wire ConnectionPool into PostgresNodeLog (or a factory) with tests that prove concurrent use works, version pin psycopg-pool, keep the license notes, or
2. Drop this PR until there is a real caller.

Do not merge unused optional deps just to look prepared. CI Integration (Postgres) passing means the existing driver still works without a pool, which undercuts the claim that this is required.

docs/THIRD_PARTY_LICENSES.md
desireddymohithreddy0925: Fair point, I've updated the PR to actually implement pooling. I refactored PostgresNodeLog to accept either a raw connection or a ConnectionPool, implemented an internal connection leasing context manager to drop the global thread lock when using a pool, and updated open_postgres_node_log to accept a pool_size argument. I also added a test_live_pool_concurrency integration test that spins up 10 threads concurrently executing claim_tool_call and append to prove thread safety under concurrency.

pyproject.toml
desireddymohithreddy0925: Fair point, I've updated the PR to actually implement pooling. I refactored PostgresNodeLog to accept either a raw connection or a ConnectionPool, implemented an internal connection leasing context manager to drop the global thread lock when using a pool, and updated open_postgres_node_log to accept a pool_size argument. I also added a test_live_pool_concurrency integration test that spins up 10 threads concurrently executing claim_tool_call and append to prove thread safety under concurrency.

.github/workflows/security-scan.yml
desireddymohithreddy0925: Fair point, I've updated the PR to actually implement pooling. I refactored PostgresNodeLog to accept either a raw connection or a ConnectionPool, implemented an internal connection leasing context manager to drop the global thread lock when using a pool, and updated open_postgres_node_log to accept a pool_size argument. I also added a test_live_pool_concurrency integration test that spins up 10 threads concurrently executing claim_tool_call and append to prove thread safety under concurrency.

PR #295, opened 2026-08-26
fix(python): add dbos importorskip to sandbox test

desireddymohithreddy0925:
### Summary
Fixes a crash during `pytest` collection when the optional `dbos` dependency is not installed. 

### Details
Previously, `conformance/r06_sandbox_test.py` imported `trajectory_client` before attempting to skip the test. Since the client adapter imports `dbos` at module level, this caused Python to crash before the `importorskip` could ever run. 

This PR:
1. Moves `pytest.importorskip("dbos")` to the top of `conformance/r06_sandbox_test.py` so it properly skips without crashing during collection.
2. Removes `dbos` from the hard `dependencies` list in `pyproject.toml` and correctly defines it as an optional dependency (`[project.optional-dependencies] dbos = ["dbos"]`), accurately reflecting that `dbos` is an optional feature.

Resolves #295

### DIFF
--- a/conformance/r06_sandbox_test.py
+++ b/conformance/r06_sandbox_test.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 import pytest
 
+pytest.importorskip("dbos")
+
 from client.python.trajectory_client import exec_tool, open_trajectory
 from trajectory_ir.effects import EffectClass
 from trajectory_ir.runtime.sandbox import RunMode, SandboxForbidden
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,12 +26,11 @@ classifiers = [
     "Programming Language :: Python :: 3.14",
     "Topic :: Software Development :: Libraries",
 ]
-dependencies = [
-    "dbos",
-]
+dependencies = []
 
 [project.optional-dependencies]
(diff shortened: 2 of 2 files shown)


Siddhartha-singh01 (changes requested):
This does not fix what the PR claims. I checked the file order and ran the module.

conformance/r06_sandbox_test.py currently does:

1. from client.python.trajectory_client import exec_tool, open_trajectory
2. later pytest.importorskip("dbos")

trajectory_client imports drivers.durable_backend.dbos.adapter at import time, and that module does from dbos import DBOS. If dbos is missing, Python dies on that import. importorskip never runs. You put the parachute on after the plane already hit the ground.

Also:

dbos is a hard dependency in pyproject.toml, not a [dbos] extra. The PR description invents an extra that does not exist.

Commit message says Resolves #294. #294 is the psycopg-pool PR. Wrong issue. Do not spray Resolves at random numbers.

What I expect before this can merge:

1. Move importorskip (or a try/except ImportError skip) ABOVE any trajectory_client / dbos adapter import.
2. Fix the PR body. Be honest about whether dbos is required or optional.
3. Fix the commit trailer to the real issue, or drop it.
4. Show a local repro: uninstall dbos, run pytest conformance/r06_sandbox_test.py, prove it skips instead of exploding on collection.

As written this is theater. CI is green only because dbos is installed in CI.

PR #296, opened 2026-08-26
docs: add trajirctl design spec

Ayush-Patel-56:
## Summary

Adds a design spec for trajirctl, a small command line tool for operating on trajectories directly from a terminal instead of only through the MCP server. Right now the only way to check status, export/import a .tir package, or verify a signature is by having an AI agent call the MCP tools. This spec plans a CLI that wraps the same Go SDK calls (go/trajir/client, go/trajir/log, go/trajir/tir) so a person can run these operations by hand.

The tool itself will live in a separate repo since it's a consumer of this module rather than part of the core library. This PR only adds the spec and calls out two small pieces of prep work needed in this repo before that CLI can depend on it cleanly:

1. Pull the workdir resolution and validation helpers that trajir-mcp's tools already use (requireBoundedWorkDir, workdirSQLitePaths in go/trajir/mcp/tools.go) into their own importable package, go/trajir/workdir, so the CLI isn't reimplementing the same path handling. trajir-mcp's behavior does not change.
2. Cut a go/vX.Y.Z tag once that extraction lands, since the Go module is in a subdirectory and needs its own tag prefix for a consumer to depend on a real version instead of a raw commit SHA. docs/RELEASE.md already covers the tagging process, this just adds the missing subdirectory tag alongside the usual one.

Neither of those two follow ups is in this PR, this is spec only.

## Test plan

- [x] Docs only change, no code touched
- [ ] N/A for unit/integration tests

## Checklist

- [x] New functionality includes tests in PR (not applicable, docs only)
- [x] Change matches master README (no invented APIs)
- [x] Relevant CI checks pass locally (or will pass on PR)
- [x] AI assistance disclosed below if tool wrote a meaningful share of the change

## Safety areas

- [x] No
- [ ] Yes (call it out here)

## AI assistance

Worked through this with Claude Code end to end, from scoping the CLI down to a plan through writing the spec doc itself. I reviewed the design and the written spec before committing.


docs/superpowers/specs/2026-08-27-trajirctl-design.md
```diff
@@ -0,0 +1,195 @@
+# trajirctl: an operator CLI for Trajectory IR
```
Siddhartha-singh01, line 1: Solid operator CLI design. Separate repo + thin SDK wrap is the path I want.

Siddhartha-singh01 (approved):
Mentor read on the trajirctl design spec.

This is the right problem to write down. MCP is for agents. Humans still need a hand tool for status, export, import, and verify without spinning up an agent chat. Keeping trajirctl in a separate repo and extracting workdir helpers here first is also the right cut. Do not grow another mini framework inside Trajectory IR for six flat commands.

I like the explicit reject of a Backend interface in v1. We have one real backend shape today. Inventing the second one in the dark is how you get an interface nobody can implement cleanly later.

Stdlib flag over Cobra matches crashagent and trajir mcp. Good.

Prep work in §2 (extract go/trajir/workdir, cut go/vX.Y.Z tags) should land as its own small PR before anyone starts the external repo against main tip. Call that out when you open the follow on so it does not get lost.

Docs only, CI green. Approved as design direction.

jakharmonika364 (approved):
LGTM

PR #306, opened 2026-08-27
fix(client): enforce strict tenant isolation in Resume

aneek22112007-tech:
## Summary
This PR fixes a severe tenant isolation flaw in `client.Resume()` where it was previously validating trajectory existence using `ListNodesAllTenants(trajectoryID)`. This allowed a caller to successfully resume a trajectory created by a different tenant, resulting in mixed-tenant cross-contamination of the IR history upon subsequent appends.

By replacing the check with the tenant-scoped `ListNodes(trajectoryID, tenantID)` method, we ensure that a trajectory can only be resumed if it is genuinely owned by the requesting tenant.

## Changes
- **`go/trajir/client/client.go`**: Switched from `ListNodesAllTenants` to `ListNodes` in `Resume()` to enforce the tenant isolation boundary natively at the read path.
- **`go/trajir/client/client_test.go`**: Added `TestResumeTenantIsolation` to explicitly assert that attempting to resume a valid trajectory ID using a mismatched tenant ID safely fails with a "no existing nodes" error.

## Validation
- Ran `go test ./...` in the `go` module. All tests passed, including the new regression test. No existing public API signatures were altered.

### DIFF
--- a/go/trajir/client/client.go
+++ b/go/trajir/client/client.go
@@ -128,7 +128,7 @@ func Resume(tenantID, trajectoryID string, opts Options) (*Trajectory, error) {
 	if err != nil {
 		return nil, err
 	}
-	rows, err := tr.log.ListNodesAllTenants(trajectoryID)
+	rows, err := tr.log.ListNodes(trajectoryID, tenantID)
 	if err != nil {
 		_ = tr.Close()
 		return nil, err
--- a/go/trajir/client/client_test.go
+++ b/go/trajir/client/client_test.go
@@ -227,3 +227,25 @@ func TestExecToolLogsToolCallOnError(t *testing.T) {
 		t.Fatalf("expected no TOOL_RESULT on error, got %v (err: %v)", hasResult, err)
 	}
 }
+
+func TestResumeTenantIsolation(t *testing.T) {
+	dir := t.TempDir()
+	opts := client.Options{WorkDir: dir}
+
+	trA, err := client.OpenTrajectory("tenant-A", "shared-traj", opts)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, err := trA.Project(1, map[string]any{"goal": "x"}); err != nil {
+		t.Fatal(err)
+	}
+	_ = trA.Close()
+
+	_, err = client.Resume("tenant-B", "shared-traj", opts)
+	if err == nil {
+		t.Fatal("expected Resume to fail for a different tenant, but it succeeded")
+	}
+	if !strings.Contains(err.Error(), "no existing nodes") {
+		t.Fatalf("expected 'no existing nodes' error, got: %v", err)
+	}
+}


go/trajir/client/client.go
```diff
@@ -128,7 +128,7 @@ func Resume(tenantID, trajectoryID string, opts Options) (*Trajectory, error) {
 	if err != nil {
 		return nil, err
 	}
-	rows, err := tr.log.ListNodesAllTenants(trajectoryID)
+	rows, err := tr.log.ListNodes(trajectoryID, tenantID)
```
Siddhartha-singh01, line 131: This is the actual fix. ListNodes is the tenant-scoped read. ListNodesAllTenants stays for Export's mixed-tenant check, which is documented on the method. Resume is not that caller.

go/trajir/client/client_test.go
```diff
@@ -227,3 +227,25 @@ func TestExecToolLogsToolCallOnError(t *testing.T) {
...
+	if _, err := trA.Project(1, map[string]any{"goal": "x"}); err != nil {
+		t.Fatal(err)
+	}
+	_ = trA.Close()
+
+	_, err = client.Resume("tenant-B", "shared-traj", opts)
+	if err == nil {
+		t.Fatal("expected Resume to fail for a different tenant, but it succeeded")
+	}
+	if !strings.Contains(err.Error(), "no existing nodes") {
```
Siddhartha-singh01, line 248: Keep matching the existing error string. If we invent a "wrong tenant" error we tell tenant-B that shared-traj exists. Same wording as a missing workdir is the right isolation behavior.

Siddhartha-singh01 (dismissed):
Looked at 8efdbf0.

Resume was using ListNodesAllTenants as an existence check. That helper is the deliberate escape hatch for tir.Export when it needs to detect mixed tenant_id values on one trajectory_id. It is not the read path for a tenant-bound client handle. If tenant-B can Resume a trajectory that only tenant-A wrote, the handle comes back with TenantID=B and the next Project/Seal/Exec appends under the wrong tenant. That is how you contaminate an IR log without ever failing a call.

Switching to ListNodes(trajectoryID, tenantID) is the correct boundary. Empty result still hits the existing "no existing nodes" error, which is what we want. We should not start saying "wrong tenant" here; that leaks that the trajectory exists.

Test is the right shape: one sqlite, two tenants, same trajectory_id, Resume must fail. CI Go job is green.

Not this PR: Python resume() in client/python/trajectory_client.py still calls list_nodes_all_tenants. Same hole on the reference port. File a follow-up, do not grow this diff.

Approving.

jakharmonika364 (changes requested):
same bug as this PR in python client, `resume()` still uses `list_nodes_all_tenants` (line ~210), should use `list_nodes(trajectory_id, tenant_id=tenant_id)` like `NodeLog` already supports at `log.py:236`

Ayush-Patel-56:
please sign off DCO `git commit -s --amend --no-edit`

Ayush-Patel-56:
hey, please make the commit history clean

Ayush4958:
@aneek22112007-tech 
could u squash ur commit
Its too messy, it will make it more clean & neat 
thanks ..

Ayush4958 (approved):
LGTM

Ayush-Patel-56:
@aneek22112007-tech can u please re-write the signoff mail ?

jakharmonika364 (approved):
LGTM, thanks!

jakharmonika364:
hey @aneek22112007-tech, refer the issue number in the PR description

PR #310, opened 2026-08-28
fix(client): scope Python resume() to the calling tenant

### LINKED ISSUE #309: bug: Python resume() ignores tenant when checking history
## What happened

Python `resume()` checks history with `NodeLog.list_nodes_all_tenants(trajectory_id)`. The tenant argument is ignored.

Go has the same hole in `client.Resume` (see #306). Even after that lands, the Python client still attaches if *any* tenant already wrote that `trajectory_id` in the same sqlite file.

After a successful resume the handle uses the *caller* tenant on later `project` / `seal_decision` / `exec_tool` calls, so you get mixed `tenant_id` rows on one trajectory.

## What you expected

`resume()` should call `list_nodes(trajectory_id, tenant_id=tenant_id)`. If that tenant has no nodes, raise the existing "no existing nodes" error. Do not advertise that another tenant already owns the id.

Close the NodeLog after the check. The one-shot constructor currently relies on `__del__`.

## Steps to reproduce

1. `open_trajectory("tenant-A", "shared-traj", db_path=db)` and write a node
2. `resume("shared-traj", tenant_id="tenant-B", db_path=db)`
3. That call succeeds today

## Environment

Python 3.12, current `main`

## Related

#306 (Go Resume)

Siddhartha-singh01:
## Summary

Python `resume()` was asking the log "does this trajectory_id exist for anyone?" via `list_nodes_all_tenants`. Tenant B could attach to tenant A's run in a shared sqlite file and then write nodes under B.

It now uses `list_nodes(..., tenant_id=tenant_id)`, same boundary as Go #306. Empty result still raises the existing "no existing nodes" error, so we do not leak that another tenant already has the id.

## Related issues

Closes #309
Related: #306

## How I tested

```bash
pytest test/unit/test_client_sdk_resume_alias.py -q
```

6 passed, including a new test that tenant-B cannot resume tenant-A's `shared-traj`.

## Checklist

## Safety areas

- [x] Yes. Client `resume()` existence check only. Gate / seal code is unchanged.

## AI assistance (if any)

Editor help on the patch. I wrote the test and ran it.

### DIFF
--- a/client/python/trajectory_client.py
+++ b/client/python/trajectory_client.py
@@ -207,7 +207,7 @@ def resume(
         mode=normalize_run_mode(mode),
     )
     try:
-        nodes = traj._log.list_nodes_all_tenants(trajectory_id)
+        nodes = traj._log.list_nodes(trajectory_id, tenant_id=tenant_id)
     except Exception:
         traj.close()
         raise
--- a/drivers/postgres/log.py
+++ b/drivers/postgres/log.py
@@ -324,6 +324,8 @@ class PostgresNodeLog:
         *,
         tenant_id: str,
     ) -> list[dict[str, Any]]:
+        if not tenant_id:
+            raise ValueError("tenant_id must be a non-empty string")
         return self._list_nodes(trajectory_id, tenant_id=tenant_id)
 
     def list_nodes_all_tenants(self, trajectory_id: str) -> list[dict[str, Any]]:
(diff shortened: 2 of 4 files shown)


Ayush-Patel-56 (approved):
Cheers!

jakharmonika364 (approved):
LGTM

PR #338, opened 2026-09-03
fix: bind live docker stack to loopback and parameterize credentials

### LINKED ISSUE #337: bug: hardcoded default credentials and wildcard host port bindings in live docker compose and matrix scripts
## What happened

`docker-compose.live.yml` hardcodes default credentials for Postgres (`trajir:trajir`), MinIO (`minioadmin:minioadmin`), and Temporal Postgres (`temporal:temporal`). In addition, the compose file maps service ports directly to all network interfaces on the host (`5432:5432`, `9000:9000`, `9001:9001`, `7233:7233`) instead of binding to loopback (`127.0.0.1`).

Similarly, `scripts/run_live_matrix.sh` and `scripts/run_live_matrix.ps1` fall back to these static credentials by default:
* `TRAJIR_DATABASE_URL` defaults to `postgresql://trajir:trajir@127.0.0.1:5432/trajir`
* `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` default to `minioadmin`

While these defaults are intended for local test execution, binding `0.0.0.0` with known static credentials exposes the database and S3 artifact store to any host on the local network (such as shared office Wi-Fi or internal subnets). If host firewalling is missing or Docker Desktop exposes forwarded ports, external entities can access or alter the database and CAS storage. Furthermore, using these files as templates for staging or shared development risks deploying insecure defaults.

### Affected locations

* `docker-compose.live.yml` (lines 15-19, 31-33, 45-49, 67-71)
* `scripts/run_live_matrix.sh` (lines 31-36)
* `scripts/run_live_matrix.ps1` (lines 16-22)
* `docs/LIVE_INTEGRATION_DOCKER.md`

## What you expected

1. Published ports in `docker-compose.live.yml` should bind explicitly to loopback (`127.0.0.1:5432:5432`, `127.0.0.1:9000:9000`, `127.0.0.1:9001:9001`, `127.0.0.1:7233:7233`) so services are not accessible over external interfaces.
2. Credentials should be configurable via environment variables or a local gitignored `.env` file rather than hardcoded in the compose definition.
3. Live matrix test scripts should warn if they run against non-loopback endpoints using default credentials.
4. Add a prominent notice in `docker-compose.live.yml` stating that the file is intended only for local development.

## Steps to reproduce

1. Inspect port definitions and environment blocks in `docker-compose.live.yml`:
```yaml
postgres:
  environment:
    POSTGRES_USER: trajir
    POSTGRES_PASSWORD: trajir
  ports:
    - "5432:5432"
```
2. Start the services on a machine connected to a local network:
```bash
docker compose -f docker-compose.live.yml up -d
```
3. Connect from another machine on the same network:
```bash
psql -h <host-ip> -U trajir -d trajir
```
The connection succeeds without needing to discover or break any credentials.

## Environment

Docker Compose v2, Linux / macOS / Windows Docker Desktop, Bash / PowerShell scripts

## Logs or output

```text
Ports bound to all interfaces:
0.0.0.0:5432->5432/tcp
0.0.0.0:9000->9000/tcp
0.0.0.0:9001->9001/tcp
0.0.0.0:7233->7233/tcp
```

## Related spec section

SECURITY.md, docs/LIVE_INTEGRATION_DOCKER.md

Krirox:
## Summary

This PR addresses security issue #337 by binding all exposed container ports in `docker-compose.live.yml` strictly to `127.0.0.1` (loopback), parameterizing credentials via environment variables with defaults, adding a `.env.example` template, updating the matrix smoke scripts to source `.env` and warn when default credentials are used with non-loopback endpoints, and documenting loopback bindings.

## Related issues

Closes #337

## Files changed

| File | Change |
|------|--------|
| `docker-compose.live.yml` | Bind published ports to `127.0.0.1`, parameterize credentials with defaults, add security warning header |
| `.env.example` | Template for local credential and endpoint overrides |
| `scripts/run_live_matrix.sh` | Source `.env`, parameterize credentials, update healthcheck, add non-loopback warning |
| `scripts/run_live_matrix.ps1` | Support `.env` loading, parameterize credentials, update healthcheck, add non-loopback warning |
| `docs/LIVE_INTEGRATION_DOCKER.md` | Document loopback bindings and `.env.example` usage |

## How I tested

```bash
# Validated compose configuration
docker compose -f docker-compose.live.yml config

# Validated shell script syntax and loopback warnings
bash -n scripts/run_live_matrix.sh
./scripts/run_live_matrix.sh --up-only --skip-up

# Validated PowerShell script syntax and loopback warnings
.\scripts\run_live_matrix.ps1 -UpOnly -SkipUp

# Python unit test suite
pytest test/unit -q  # 190 passed, 2 skipped

# Go test suite
cd go && go test ./... -count=1  # all packages passed
```

## Checklist

- [x] If you changed `.github/workflows/**`, note it under Safety and expect extra review

## Safety areas

- [x] No
- [ ] Yes (call it out here)

## AI assistance (if any)

None.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,6 @@
 POSTGRES_USER=trajir
 POSTGRES_PASSWORD=trajir
 POSTGRES_DB=trajir
-TRAJIR_DATABASE_URL=postgresql://trajir:trajir@127.0.0.1:5432/trajir
 
 # MinIO (S3 CAS)
 MINIO_ROOT_USER=minioadmin
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,25 @@
+# Configuration template for local development integration testing.
+# Copy this file to .env to override defaults when using docker-compose.live.yml
+# and scripts/run_live_matrix.sh / .ps1.
+# Note: .env is ignored by git.
+
+# Postgres (NodeLog)
+POSTGRES_USER=trajir
+POSTGRES_PASSWORD=trajir
+POSTGRES_DB=trajir
+TRAJIR_DATABASE_URL=postgresql://trajir:trajir@127.0.0.1:5432/trajir
+
+# MinIO (S3 CAS)
+MINIO_ROOT_USER=minioadmin
+MINIO_ROOT_PASSWORD=minioadmin
+TRAJIR_S3_ENDPOINT_URL=http://127.0.0.1:9000
+TRAJIR_S3_BUCKET=trajir
+AWS_ACCESS_KEY_ID=minioadmin
+AWS_SECRET_ACCESS_KEY=minioadmin
+AWS_DEFAULT_REGION=us-east-1
+
+# Temporal (Go durable backend)
+TEMPORAL_POSTGRES_USER=temporal
+TEMPORAL_POSTGRES_PASSWORD=temporal
+TEMPORAL_POSTGRES_DB=temporal
+TEMPORAL_HOSTPORT=localhost:7233
--- a/.env.example
+++ b/.env.example
@@ -7,14 +7,19 @@
 POSTGRES_USER=trajir
 POSTGRES_PASSWORD=trajir
 POSTGRES_DB=trajir
+# Note: TRAJIR_DATABASE_URL is automatically derived from POSTGRES_* by run_live_matrix scripts:
+# TRAJIR_DATABASE_URL=postgresql://<user>:<password>@127.0.0.1:5432/<db>
+# (Components are URL-encoded by the scripts)
 
 # MinIO (S3 CAS)
 MINIO_ROOT_USER=minioadmin
 MINIO_ROOT_PASSWORD=minioadmin
 TRAJIR_S3_ENDPOINT_URL=http://127.0.0.1:9000
 TRAJIR_S3_BUCKET=trajir
-AWS_ACCESS_KEY_ID=minioadmin
-AWS_SECRET_ACCESS_KEY=minioadmin
+# AWS credentials default to MINIO_ROOT_USER and MINIO_ROOT_PASSWORD.
+# scripts/run_live_matrix automatically sets them if not explicitly provided.
+# AWS_ACCESS_KEY_ID=minioadmin
+# AWS_SECRET_ACCESS_KEY=minioadmin
 AWS_DEFAULT_REGION=us-east-1
 
 # Temporal (Go durable backend)
--- a/.env.example
+++ b/.env.example
@@ -26,4 +26,6 @@ AWS_DEFAULT_REGION=us-east-1
 TEMPORAL_POSTGRES_USER=temporal
 TEMPORAL_POSTGRES_PASSWORD=temporal
 TEMPORAL_POSTGRES_DB=temporal
+# Optional: customize visibility database name (defaults to temporal_visibility)
+# TEMPORAL_POSTGRES_VISIBILITY_DB=temporal_visibility
 TEMPORAL_HOSTPORT=localhost:7233
--- a/docker-compose.live.yml
+++ b/docker-compose.live.yml
@@ -1,3 +1,9 @@
+# WARNING: FOR ISOLATED LOCAL DEVELOPMENT AND TESTING ONLY.
+# Published ports are bound strictly to 127.0.0.1 (loopback) to prevent
+# exposure to external networks.
+# Never expose this stack to untrusted networks or use these default
+# credentials in staging or production environments.
+#
 # Vacant live-integration sandbox for Trajectory IR maintainers.
 # Hosts Postgres (NodeLog), MinIO (S3 CAS), and Temporal (Go durable).
 #
@@ -12,13 +18,13 @@ services:
     image: postgres:16.6
     container_name: trajir-live-postgres
     environment:
-      POSTGRES_USER: trajir
-      POSTGRES_PASSWORD: trajir
-      POSTGRES_DB: trajir
+      POSTGRES_USER: ${POSTGRES_USER:-trajir}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-trajir}
+      POSTGRES_DB: ${POSTGRES_DB:-trajir}
     ports:
-      - "5432:5432"
+      - "127.0.0.1:5432:5432"
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U trajir -d trajir"]
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-trajir} -d ${POSTGRES_DB:-trajir}"]
       interval: 3s
       timeout: 5s
       retries: 30
@@ -28,11 +34,11 @@ services:
     image: postgres:16.6
     container_name: trajir-live-temporal-pg
     environment:
-      POSTGRES_USER: temporal
-      POSTGRES_PASSWORD: temporal
-      POSTGRES_DB: temporal
+      POSTGRES_USER: ${TEMPORAL_POSTGRES_USER:-temporal}
+      POSTGRES_PASSWORD: ${TEMPORAL_POSTGRES_PASSWORD:-temporal}
+      POSTGRES_DB: ${TEMPORAL_POSTGRES_DB:-temporal}
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U temporal -d temporal"]
+      test: ["CMD-SHELL", "pg_isready -U ${TEMPORAL_POSTGRES_USER:-temporal} -d ${TEMPORAL_POSTGRES_DB:-temporal}"]
       interval: 3s
       timeout: 5s
       retries: 30
@@ -42,11 +48,11 @@ services:
     container_name: trajir-live-minio
     command: server /data --console-address ":9001"
     environment:
-      MINIO_ROOT_USER: minioadmin
-      MINIO_ROOT_PASSWORD: minioadmin
+      MINIO_ROOT_USER: ${MINIO_ROOT_USER:-minioadmin}
+      MINIO_ROOT_PASSWORD: ${MINIO_ROOT_PASSWORD:-minioadmin}
     ports:
-      - "9000:9000"
-      - "9001:9001"
+      - "127.0.0.1:9000:9000"
+      - "127.0.0.1:9001:9001"
     healthcheck:
       # Official MinIO image ships curl; live endpoint is the supported probe.
       test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
@@ -64,8 +70,8 @@ services:
     environment:
       DB: postgres12
       DB_PORT: 5432
-      POSTGRES_USER: temporal
-      POSTGRES_PWD: temporal
+      POSTGRES_USER: ${TEMPORAL_POSTGRES_USER:-temporal}
+      POSTGRES_PWD: ${TEMPORAL_POSTGRES_PASSWORD:-temporal}
       POSTGRES_SEEDS: temporal-postgres
     ports:
-      - "7233:7233"
+      - "127.0.0.1:7233:7233"
--- a/docker-compose.live.yml
+++ b/docker-compose.live.yml
@@ -73,5 +73,7 @@ services:
       POSTGRES_USER: ${TEMPORAL_POSTGRES_USER:-temporal}
       POSTGRES_PWD: ${TEMPORAL_POSTGRES_PASSWORD:-temporal}
       POSTGRES_SEEDS: temporal-postgres
+      DBNAME: ${TEMPORAL_POSTGRES_DB:-temporal}
+      POSTGRES_DB: ${TEMPORAL_POSTGRES_DB:-temporal}
     ports:
       - "127.0.0.1:7233:7233"
--- a/docker-compose.live.yml
+++ b/docker-compose.live.yml
@@ -24,7 +24,7 @@ services:
     ports:
       - "127.0.0.1:5432:5432"
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-trajir} -d ${POSTGRES_DB:-trajir}"]
+      test: ["CMD", "pg_isready", "-U", "${POSTGRES_USER:-trajir}", "-d", "${POSTGRES_DB:-trajir}"]
       interval: 3s
       timeout: 5s
       retries: 30
@@ -38,7 +38,7 @@ services:
       POSTGRES_PASSWORD: ${TEMPORAL_POSTGRES_PASSWORD:-temporal}
       POSTGRES_DB: ${TEMPORAL_POSTGRES_DB:-temporal}
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U ${TEMPORAL_POSTGRES_USER:-temporal} -d ${TEMPORAL_POSTGRES_DB:-temporal}"]
+      test: ["CMD", "pg_isready", "-U", "${TEMPORAL_POSTGRES_USER:-temporal}", "-d", "${TEMPORAL_POSTGRES_DB:-temporal}"]
       interval: 3s
       timeout: 5s
       retries: 30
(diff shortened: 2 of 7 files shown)


scripts/run_live_matrix.sh
```diff
@@ -28,14 +28,42 @@ for arg in "$@"; do
...
 export AWS_DEFAULT_REGION="${AWS_DEFAULT_REGION:-us-east-1}"
 export TEMPORAL_HOSTPORT="${TEMPORAL_HOSTPORT:-localhost:7233}"
 
+is_loopback() {
+  local target="$1"
+  if [[ "$target" == *127.0.0.1* || "$target" == *localhost* || "$target" == *::1* ]]; then
+    return 0
+  fi
+  return 1
+}
```
jakharmonika364, line 108: Substring match, not anchored, a host like `mylocalhost-proxy.example.com` or `IPv6 2001:db8::123` matches this and skips the warning. Anchor it (`^127\.`, exact `localhost`, `^\[::1\]$` etc).

scripts/run_live_matrix.ps1
```diff
@@ -13,14 +13,44 @@ $ErrorActionPreference = "Stop"
...
-$env:AWS_ACCESS_KEY_ID = if ($env:AWS_ACCESS_KEY_ID) { $env:AWS_ACCESS_KEY_ID } else { "minioadmin" }
-$env:AWS_SECRET_ACCESS_KEY = if ($env:AWS_SECRET_ACCESS_KEY) { $env:AWS_SECRET_ACCESS_KEY } else { "minioadmin" }
+$env:AWS_ACCESS_KEY_ID = if ($env:AWS_ACCESS_KEY_ID) { $env:AWS_ACCESS_KEY_ID } else { $minioUser }
+$env:AWS_SECRET_ACCESS_KEY = if ($env:AWS_SECRET_ACCESS_KEY) { $env:AWS_SECRET_ACCESS_KEY } else { $minioPassword }
 $env:AWS_DEFAULT_REGION = if ($env:AWS_DEFAULT_REGION) { $env:AWS_DEFAULT_REGION } else { "us-east-1" }
 $env:TEMPORAL_HOSTPORT = if ($env:TEMPORAL_HOSTPORT) { $env:TEMPORAL_HOSTPORT } else { "localhost:7233" }
 
+function Test-IsLoopback($target) {
+    return ($target -match "127\.0\.0\.1|localhost|\[::1\]")
+}
```
jakharmonika364, line 62: Same substring issue as the bash version, `-match "localhost"` isn't anchored.

scripts/run_live_matrix.sh
```diff
@@ -28,14 +28,42 @@ for arg in "$@"; do
   esac
 done
 
-export TRAJIR_DATABASE_URL="${TRAJIR_DATABASE_URL:-postgresql://trajir:trajir@127.0.0.1:5432/trajir}"
+if [[ -f .env ]]; then
+  # shellcheck disable=SC1091
+  set -a
+  source .env
+  set +a
```
jakharmonika364: `source` shell-evals the values, `$`, backticks, `$()` in a password get expanded/executed. Parse it as literal KEY=VALUE like the PS1 version does.

.env.example
```diff
@@ -0,0 +1,25 @@
+# Configuration template for local development integration testing.
+# Copy this file to .env to override defaults when using docker-compose.live.yml
+# and scripts/run_live_matrix.sh / .ps1.
+# Note: .env is ignored by git.
+
+# Postgres (NodeLog)
+POSTGRES_USER=trajir
+POSTGRES_PASSWORD=trajir
+POSTGRES_DB=trajir
+TRAJIR_DATABASE_URL=postgresql://trajir:trajir@127.0.0.1:5432/trajir
```
jakharmonika364: Hardcoding this next to `POSTGRES_PASSWORD` means changing the password alone silently breaks the URL. Either derive it always or drop it from the example.

docker-compose.live.yml
```diff
@@ -64,8 +70,8 @@ services:
     environment:
       DB: postgres12
       DB_PORT: 5432
-      POSTGRES_USER: temporal
-      POSTGRES_PWD: temporal
+      POSTGRES_USER: ${TEMPORAL_POSTGRES_USER:-temporal}
+      POSTGRES_PWD: ${TEMPORAL_POSTGRES_PASSWORD:-temporal}
```
jakharmonika364, line 74: `TEMPORAL_POSTGRES_DB` isn't passed here, setting it in `.env` renames `temporal-postgres` but temporal still points at the default DB name.

jakharmonika364 (changes requested):

scripts/run_live_matrix.sh
```diff
@@ -57,8 +103,9 @@ if [[ "$SKIP_UP" -eq 0 ]]; then
   done
 
   echo "==> wait minio healthy"
+  minio_live_url="${TRAJIR_S3_ENDPOINT_URL%/}/minio/health/live"
```
Siddhartha-singh01: **[bug] Framework breakage: MinIO health wait must not use `TRAJIR_S3_ENDPOINT_URL`.**

This couples local container readiness to the client endpoint override. If someone sets `TRAJIR_S3_ENDPOINT_URL` to a remote/custom S3 endpoint (exactly what `.env.example` now invites), this loop probes the wrong host, can hang for ~2 minutes, then fail with `minio not ready` even when `trajir-live-minio` is healthy.

Keep the readiness probe on the published local port, e.g.:

```bash
minio_live_url="http://127.0.0.1:9000/minio/health/live"
```

Use `TRAJIR_S3_ENDPOINT_URL` only for the later client/test calls, not for compose service readiness.

scripts/run_live_matrix.ps1
```diff
@@ -33,17 +64,18 @@ if (-not $SkipUp) {
...
-        docker exec trajir-live-postgres pg_isready -U trajir -d trajir 2>$null | Out-Null
+        docker exec trajir-live-postgres pg_isready -U $postgresUser -d $postgresDb 2>$null | Out-Null
         if ($LASTEXITCODE -eq 0) { $ok = $true; break }
         Start-Sleep -Seconds 2
     }
     if (-not $ok) { throw "postgres not ready" }
 
     Write-Host "==> wait minio healthy"
     $ok = $false
+    $minioLiveUrl = "$($env:TRAJIR_S3_ENDPOINT_URL.TrimEnd('/'))/minio/health/live"
```
Siddhartha-singh01: **[bug] Same MinIO readiness coupling on Windows.**

`` is derived from `TRAJIR_S3_ENDPOINT_URL`. That breaks `-UpOnly` / full matrix runs whenever the S3 endpoint override is not the local MinIO listener.

Probe `http://127.0.0.1:9000/minio/health/live` (or another fixed loopback URL matching the compose publish), and keep `TRAJIR_S3_ENDPOINT_URL` for boto/AWS client usage only.

.env.example
```diff
@@ -0,0 +1,24 @@
...
+POSTGRES_PASSWORD=trajir
+POSTGRES_DB=trajir
+
+# MinIO (S3 CAS)
+MINIO_ROOT_USER=minioadmin
+MINIO_ROOT_PASSWORD=minioadmin
+TRAJIR_S3_ENDPOINT_URL=http://127.0.0.1:9000
+TRAJIR_S3_BUCKET=trajir
+AWS_ACCESS_KEY_ID=minioadmin
+AWS_SECRET_ACCESS_KEY=minioadmin
```
Siddhartha-singh01: **[bug] Credential desync footgun (same class as the earlier `TRAJIR_DATABASE_URL` review comment).**

`.env.example` sets both:

- `MINIO_ROOT_USER` / `MINIO_ROOT_PASSWORD` (used by compose to start MinIO)
- `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` (used by clients)

Scripts only fall back to MinIO root creds when `AWS_*` are unset. Because these are explicitly set here, changing only `MINIO_ROOT_PASSWORD` leaves clients on the old secret and all MinIO live tests fail.

Please remove `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` from `.env.example` and let the scripts derive them from `MINIO_ROOT_*`, or add a loud comment plus script-side validation that they match.

scripts/run_live_matrix.sh
```diff
@@ -28,14 +28,60 @@ for arg in "$@"; do
...
+  done < .env
+fi
+
+POSTGRES_USER="${POSTGRES_USER:-trajir}"
+POSTGRES_PASSWORD="${POSTGRES_PASSWORD:-trajir}"
+POSTGRES_DB="${POSTGRES_DB:-trajir}"
+MINIO_ROOT_USER="${MINIO_ROOT_USER:-minioadmin}"
+MINIO_ROOT_PASSWORD="${MINIO_ROOT_PASSWORD:-minioadmin}"
+
+export TRAJIR_DATABASE_URL="${TRAJIR_DATABASE_URL:-postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@127.0.0.1:5432/${POSTGRES_DB}}"
```
Siddhartha-singh01: **[bug] Unencoded credentials in DSN will break customized passwords.**

```bash
postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@127.0.0.1:5432/${POSTGRES_DB}
```

If `POSTGRES_PASSWORD` contains `@`, `:`, `/`, `#`, `?`, or `%`, the URL parses incorrectly and every Postgres live test fails. This PR is specifically asking people to customize credentials via `.env`, so this will get hit.

URL-encode user and password (e.g. `printf %s "" | jq -sRr @uri`, or a tiny Python one-liner) before building `TRAJIR_DATABASE_URL`. Apply the same fix in the PowerShell script.

docker-compose.live.yml
```diff
@@ -12,13 +18,13 @@ services:
...
-      POSTGRES_DB: trajir
+      POSTGRES_USER: ${POSTGRES_USER:-trajir}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-trajir}
+      POSTGRES_DB: ${POSTGRES_DB:-trajir}
     ports:
-      - "5432:5432"
+      - "127.0.0.1:5432:5432"
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U trajir -d trajir"]
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-trajir} -d ${POSTGRES_DB:-trajir}"]
```
Siddhartha-singh01: **[bug/security] Unquoted interpolation into `CMD-SHELL`.**

```yaml
test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-trajir} -d ${POSTGRES_DB:-trajir}"]
```

Compose substitutes these before the container runs. A username/db value with spaces or shell metacharacters becomes healthcheck shell injection / broken probes. Same pattern on the Temporal Postgres healthcheck below.

For this sandbox, either:

1. Keep healthchecks on fixed literals (`trajir` / `temporal`) and document that healthchecks assume defaults, or
2. Quote safely, e.g. `pg_isready -U ""${POSTGRES_USER:-trajir}"" -d ""${POSTGRES_DB:-trajir}""`, and validate that user/db are simple identifiers in the smoke scripts.

Given this PR is a security hardening change, please do not leave unquoted shell interpolation in healthchecks.

docker-compose.live.yml
```diff
@@ -64,8 +70,10 @@ services:
     environment:
       DB: postgres12
       DB_PORT: 5432
-      POSTGRES_USER: temporal
-      POSTGRES_PWD: temporal
+      POSTGRES_USER: ${TEMPORAL_POSTGRES_USER:-temporal}
+      POSTGRES_PWD: ${TEMPORAL_POSTGRES_PASSWORD:-temporal}
       POSTGRES_SEEDS: temporal-postgres
+      DBNAME: ${TEMPORAL_POSTGRES_DB:-temporal}
```
Siddhartha-singh01, line 76: **[nit] `DBNAME` is the right Temporal auto-setup knob (good fix).**

`POSTGRES_DB` on the `temporal` service is unused by `temporalio/auto-setup` (it looks at `DBNAME` / `VISIBILITY_DBNAME`). Harmless, but consider dropping `POSTGRES_DB` here to avoid implying it does something.

Also consider setting `VISIBILITY_DBNAME` explicitly if `TEMPORAL_POSTGRES_DB` customization is supported, so visibility DB naming stays intentional rather than implicit.

Siddhartha-singh01 (changes requested):
## Review summary

Thanks for the loopback binding and credential parameterization work. The security intent is right, and the follow-up commit fixed the earlier `source` / substring-match / Temporal `DBNAME` issues.

I am still requesting changes because a few remaining paths can break the live-integration framework once people actually customize credentials or endpoints via `.env` (which this PR encourages).

### Blocking

1. **MinIO readiness probe is coupled to `TRAJIR_S3_ENDPOINT_URL`**  stack health should probe the local published MinIO port, not the client endpoint env var.
2. **`.env.example` still hardcodes `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` next to `MINIO_ROOT_*`**  changing MinIO root creds alone silently desyncs client auth (same class of bug as the earlier `TRAJIR_DATABASE_URL` comment).
3. **Constructed `TRAJIR_DATABASE_URL` is not URL-encoded** customized passwords containing `@ : / # ?` break Postgres live tests.
4. **Compose healthchecks interpolate user/db into `CMD-SHELL` unquoted**  metacharacters become shell injection in the healthcheck.

### Non-blocking but please address

5. Checklist claims tests were added; there are still no tests for `.env` parsing or loopback helpers.
6. Docs Environment section still hardcodes default DSNs without stating that `TRAJIR_DATABASE_URL` / `AWS_*` must stay consistent with `POSTGRES_*` / `MINIO_ROOT_*` when overridden.

Please fix the blocking items before merge.

Siddhartha-singh01 (dismissed):
## Re-review after follow-up commits

Previous blocking feedback has been addressed on head `3bada40`:

1. **MinIO readiness** now probes fixed `http://127.0.0.1:9000/minio/health/live` in both shell and PowerShell (no longer coupled to `TRAJIR_S3_ENDPOINT_URL`).
2. **`.env.example`** comments out `AWS_*` and documents derivation from `MINIO_ROOT_*`.
3. **DSN construction** URL-encodes user/password/db (bash `urlencode` + PowerShell `EscapeDataString`).
4. **Compose healthchecks** use exec-form `CMD` arrays instead of unquoted `CMD-SHELL`.
5. **Tests** added in `test/unit/test_live_matrix_scripts.py`; docs cover credential sync.

### Remaining nits (non-blocking)
- `POSTGRES_DB` on the `temporal` service is still unused by `temporalio/auto-setup` (`DBNAME` is the real knob). Harmless; can drop later.
- Optional: set `VISIBILITY_DBNAME` explicitly if DB rename is a supported customization path.

### Verdict
**Approved.** Thanks for the thorough follow-up.

Ayush4958 (dismissed):
LGTM ,
Non-blocker mentioned by @Siddhartha-singh01 has been fixed

scripts/run_live_matrix.sh
```diff
@@ -28,14 +28,92 @@ for arg in "$@"; do
...
+    [[ -z "$line" || "$line" == \#* ]] && continue
+    if [[ "$line" == *"="* ]]; then
+      key="${line%%=*}"
+      val="${line#*=}"
+      key="${key#"${key%%[![:space:]]*}"}"
+      key="${key%"${key##*[![:space:]]}"}"
+      val="${val#"${val%%[![:space:]]*}"}"
+      val="${val%"${val##*[![:space:]]}"}"
+      if [[ ("$val" == \"*\" && "$val" == *\" && ${#val} -ge 2) || ("$val" == \'*\' && "$val" == *\' && ${#val} -ge 2) ]]; then
+        val="${val:1:-1}"
```
jakharmonika364:
```suggestion
        val="${val:1}"; val="${val%?}"
```
`${val:1:-1}` needs bash `4.2+` for the negative index. `macOS` ships bash `3.2` as `/bin/bash`, so this'll hard-fail on any quoted `.env` value there (set `-e`). I think the suggested fix works everywhere.

jakharmonika364 (changes requested):
Everything else looks solid and ready.

jakharmonika364 (approved):
after too much iterations, this is now good to bump. LGTM

Siddhartha-singh01 (approved):
Took another look after the follow ups.

Loopback binding is in place, MinIO health hits 127.0.0.1, AWS keys are commented in .env.example so they stay in sync with MinIO, DSNs get URL encoded, and the healthchecks use CMD lists. Tests and docs look fine. CI is green.

Merging this one.

PR #355, opened 2026-09-10
fix(mcp): close TOCTOU race in path confinement (CWE-367)

### LINKED ISSUE #354: bug: MCP path confinement has a validate-then-reopen race (CWE-367) in trajectory_import_tir and trajectory_verify_signature
## What happened

In `go/trajir/mcp/tools.go`, both `toolImportTIR` and `toolVerifySignature` check the incoming path using `requireBoundedPath(in.Path, "")` to enforce `TRAJIR_MCP_ROOT` confinement. However, after validating the path, both tools pass the raw path string to `tir.Load(path)` and `tir.Verify(path)`.

Inside `tir.Load` and `tir.Verify` (in `go/trajir/tir/tir.go`), `loadImpl` performs a separate `os.Stat(path)` and `zip.OpenReader(path)` from the filesystem.

This introduces a time-of-check to time-of-use (TOCTOU / CWE-367) window. If a process replaces the validated file with a symlink pointing outside `TRAJIR_MCP_ROOT` between the `requireBoundedPath` check and the `zip.OpenReader` call, `tir.Load` opens and parses an archive located outside the confined workspace root.

`go/trajir/tir/tir.go` explicitly warns about this exact vulnerability in signature checking:
`// Use the already-open archive only - never re-open path (TOCTOU / CWE-367).`
However, the MCP tool layer reintroduces the path reopen pattern.

### Affected locations

* `go/trajir/mcp/tools.go` (lines 181-185, 227-231)
* `go/trajir/tir/tir.go` (lines 530-540)

## What you expected

Path validation and file opening should happen without a reopen race. Once `requireBoundedPath` (or a helper) opens the target file under `TRAJIR_MCP_ROOT`, `tir.Load` and `tir.Verify` should accept an existing `*os.File` or `*zip.Reader` instead of reopening the path by name.

## Steps to reproduce

1. Start `trajir-mcp` with `TRAJIR_MCP_ROOT=/workspace`.
2. Prepare a valid `.tir` archive at `/workspace/package.tir`.
3. Call `trajectory_import_tir` with `{"path": "package.tir"}` while concurrently replacing `/workspace/package.tir` with a symlink to an out-of-root archive `/etc/sensitive.tir` immediately after `requireBoundedPath` resolves.
4. `tir.Load` follows the symlink during `zip.OpenReader` and loads metadata (manifest, node count, tenant ID) from the out-of-root package.

## Environment

Go 1.22+, all platforms, trajir-mcp server

## Related spec section

docs/INTEGRATIONS.md (MCP server security and path confinement), CWE-367, CWE-73

Krirox:
## Summary

	oolImportTIR and 	oolVerifySignature called 
equireBoundedPath to validate the .tir path, then passed the path string to 	ir.Load / 	ir.Verify which re-opened the file independently via os.Stat + zip.OpenReader. Between the check and the open an attacker who could replace the file (or swap a symlink) could bypass MCP root confinement.

## Root Cause

Classic validate-then-reopen race (CWE-367). Two separate filesystem opens on the same path with no guarantee the inode stays the same between calls.

## Fix

**openBoundedTIR** validates and opens the file in a single step, returning an *os.File whose identity is the same file that passed confinement. The fd is then passed to new reader-based APIs so no second open happens.

### tir.go
- Extract loadFromZip(*zip.Reader, bool) from loadImpl (internal refactor, zero behavior change for existing callers).
- Add LoadReader(io.ReaderAt, int64) public API that creates a *zip.Reader from the caller's handle.

### signature.go
- Add VerifyReader(io.ReaderAt, int64, VerifyOptions) that creates a *zip.Reader and delegates to the existing erifySignatureFromZipFiles.

### tools.go
- Add openBoundedTIR helper (validate + open in one step).
- 	oolImportTIR uses openBoundedTIR + 	ir.LoadReader.
- 	oolVerifySignature uses openBoundedTIR + 	ir.VerifyReader.

## Testing

All 16 packages pass (go test ./...), including existing MCP integration tests that exercise import, verify, path escape rejection, signed/unsigned, and tampered packages.

Fixes #354

Signed-off-by: Krirox <agarwalkrishiv06@gmail.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -243,7 +243,7 @@ jobs:
       - name: govulncheck
         working-directory: go
         run: |
-          go install golang.org/x/vuln/cmd/govulncheck@latest
+          go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
           govulncheck ./...
 
   # ---- Deep gate: required, allowed to run slower ----
--- a/go/trajir/mcp/paths_test.go
+++ b/go/trajir/mcp/paths_test.go
@@ -192,3 +192,61 @@ func TestRequireBoundedPathFailsClosedOnInvalidPreferRoot(t *testing.T) {
 		t.Fatal("expected requireBoundedPath to fail when preferRoot cannot be resolved")
 	}
 }
+
+func TestOpenBoundedTIR(t *testing.T) {
+	root := t.TempDir()
+	t.Setenv(EnvWorkspaceRoot, root)
+
+	// Valid relative file under root
+	targetFile := filepath.Join(root, "valid.tir")
+	if err := os.WriteFile(targetFile, []byte("valid content"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	f, err := openBoundedTIR("valid.tir")
+	if err != nil {
+		t.Fatalf("openBoundedTIR relative: %v", err)
+	}
+	f.Close()
+
+	// Valid absolute path under root
+	f, err = openBoundedTIR(targetFile)
+	if err != nil {
+		t.Fatalf("openBoundedTIR absolute: %v", err)
+	}
+	f.Close()
+
+	// Outside file must fail
+	outside := t.TempDir()
+	outsideFile := filepath.Join(outside, "secret.tir")
+	if err := os.WriteFile(outsideFile, []byte("secret"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := openBoundedTIR(outsideFile); err == nil {
+		t.Fatal("expected outside file to fail")
+	}
+
+	// Relative traversal must fail
+	if _, err := openBoundedTIR("../secret.tir"); err == nil {
+		t.Fatal("expected .. to fail")
+	}
+
+	// Directory must fail
+	if _, err := openBoundedTIR("."); err == nil {
+		t.Fatal("expected directory to fail")
+	}
+
+	// Empty path must fail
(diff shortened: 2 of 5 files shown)


Ayush-Patel-56:
Hi @Krirox, CI is failing, please fix it

Ayush4958 (changes requested):
I am requesting changes because the gateway function `openBoundedTIR` does not actually close the primary CWE-367 vulnerability; it just moves it.

### Blocking

**`go/trajir/mcp/tools.go` (Lines 17-31 in `openBoundedTIR`)**
```go
func openBoundedTIR(rawPath string) (*os.File, error) {
	path, err := requireBoundedPath(rawPath, "") // <--- Time of Check
	if err != nil {
		return nil, err
	}
	f, err := os.Open(path) // <--- Time of Use
```
While this puts the check and the open adjacent to each other, they are still two separate OS operations using a string path. In the microsecond window between `requireBoundedPath` returning and `os.Open` executing, an attacker can still swap the file or parent directory with a malicious symlink pointing outside the boundary. `os.Open` will blindly follow it.

To properly close a TOCTOU race in Go path confinement, you cannot rely on sequential string based validation and opening. 

Please rethink the implementation of `openBoundedTIR` to securely obtain the file descriptor. The rest of the PR is structurally sound!

jakharmonika364 (approved):
`os.OpenRoot` closes the TOCTOU gap and the symlink-escape test covers it. LGTM

Ayush4958 (approved):
LGTM

PR #364, opened 2026-09-11
Reject empty trust store in verify_package instead of skipping the check

### LINKED ISSUE #314: verify_package treats an empty trust store as "trust everyone"
## Component:
`pkg/trajectory_ir/package/signature.py` `verify_package()` is supposed to fail closed when given a trust store, but an empty list is falsy in Python, so the check gets skipped entirely instead of rejecting everything.

### signature.py:266-268
```
if trusted_keys and not any(k == pub_raw for k in trusted_keys):
    raise TirSignatureError("public key not in trust store")
if trusted_key_ids and kid not in trusted_key_ids:
    raise TirSignatureError(f"key_id {kid!r} not in trust store")
```

### Steps to reproduce
1. Sign any `.tir` package with a throwaway Ed25519 key.
2. Call `verify_package("pkg.tir", require_signature=True, trusted_keys=[])`.

### Expected
raises `TirSignatureError: public key not in trust store` since `trusted_keys=[]` means "no key is trusted."

### Actual
returns a valid `SignatureInfo`. Any signer passes.

### Suggested fix 
check `trusted_keys is not None` instead of truthiness (same for `trusted_key_ids`).

Ayush-Patel-56:
Closes #314

verify_package() checks trusted_keys and trusted_key_ids with plain truthiness (`if trusted_keys`), and an empty list is falsy in Python. That means calling verify_package(..., trusted_keys=[]) skips the trust check entirely and accepts a signature from any key, when an empty list should mean "no key is trusted."

Changed both checks to `is not None` so:
- `trusted_keys=None` (default) keeps meaning no restriction
- `trusted_keys=[]` now correctly rejects every signer
- a non-empty list still needs an exact key match, same as before

Added a regression test covering the empty-list case for both trusted_keys and trusted_key_ids, plus a sanity check that a store containing the actual signer's key still verifies fine.

The Go implementation (go/trajir/tir/signature.go) already treats an empty TrustedKeys slice as "no restriction" on purpose, since Go slices have no nil vs empty distinction in that API, so no change needed there. This was Python-specific.

Ran the full signature test suite locally, all passing.

### DIFF
--- a/pkg/trajectory_ir/package/signature.py
+++ b/pkg/trajectory_ir/package/signature.py
@@ -263,9 +263,9 @@ def _verify_signature_bytes(
     if signer.get("key_id") and kid != key_id(pub_raw):
         raise TirSignatureError("signer.key_id does not match public_key")
 
-    if trusted_keys and not any(k == pub_raw for k in trusted_keys):
+    if trusted_keys is not None and not any(k == pub_raw for k in trusted_keys):
         raise TirSignatureError("public key not in trust store")
-    if trusted_key_ids and kid not in trusted_key_ids:
+    if trusted_key_ids is not None and kid not in trusted_key_ids:
         raise TirSignatureError(f"key_id {kid!r} not in trust store")
 
     return SignatureInfo(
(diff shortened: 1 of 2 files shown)


jakharmonika364 (approved):
LGTM, quick bumping it

PR #365, opened 2026-09-11
Add Redacted option to Go tir.Export for parity with Python

### LINKED ISSUE #315: Go .tir export has no redaction option
## Component: 
`go/trajir/tir/tir.go` `ExportOptions` has no `Redacted` field. `Export()` always writes `"redacted": false` into the manifest (line 376), and `go/trajir/redact` is never imported outside its own test file.

```
// tir.go:100-111, no Redacted field exists
type ExportOptions struct {
    Mode          Mode
    TenantID      *string
    Artifacts     []ArtifactRef
    ArtifactBytes map[string][]byte
    CAS           cas.Store
    SignKey       ed25519.PrivateKey
    SignerMeta    SignerMeta
}
```

### Steps to reproduce
1. In Python: `export_tir(node_log, "traj-1", "out-py.tir", redacted=True)` on a trajectory with a THOUGHT node and a secret-shaped field. Unzip, `nodes.ndjson` has THOUGHT collapsed to `{"redacted": true}`.
2. In Go: `tir.Export(nodeLog, "traj-1", "out-go.tir", tir.ExportOptions{Mode: "thin"})`. There is no field to request redaction.
3. Unzip `out-go.tir`, inspect `nodes.ndjson`.

### Expected: 
- Go export offers the same `redacted` capability as Python, since the two SDKs are documented as parity implementations of the same format.

### Actual: 
- `manifest.json` always says `"redacted": false`, THOUGHT payloads and secret-shaped fields ship raw regardless of caller intent.

### Suggested fix: 
- add `Redacted bool` to `ExportOptions`, wire it to `go/trajir/redact` before writing `nodes.ndjson`.

Ayush-Patel-56:
Closes #315

The Go ExportOptions struct had no field to request redaction, so exported packages always shipped raw THOUGHT payloads and secret-shaped fields, even though go/trajir/redact already implements the same scrubbing logic used in projection and was never wired into export. manifest.json also hardcoded "redacted": false regardless of caller intent.

Added a Redacted bool to ExportOptions. When set:
- every node is run through redact.RedactPayload before export (THOUGHT payloads collapse to {"redacted": true}, keys/values matching secret patterns become [REDACTED])
- payload_hash and id are recomputed from the redacted payload since it changes the content being hashed
- fat mode is forced down to thin and artifacts/artifact bytes are dropped, since embedded artifact bytes can hold secrets too, matching what export_tir(redacted=True) does in Python
- manifest.json now reports the real redacted value instead of a hardcoded false

Added two tests: one exporting a trajectory with a THOUGHT node and an api_key field and checking both get scrubbed and the manifest flag flips to true, and one confirming fat mode gets forced to thin with an empty artifacts manifest when redaction is requested.

Ran go build, go vet, and go test ./... across the whole module, all passing.

### DIFF
--- a/go/trajir/tir/tir.go
+++ b/go/trajir/tir/tir.go
@@ -33,6 +33,7 @@ import (
 	"github.com/Coder-s-OG-s/Trajectory-IR/go/trajir/cas"
 	nodelog "github.com/Coder-s-OG-s/Trajectory-IR/go/trajir/log"
 	"github.com/Coder-s-OG-s/Trajectory-IR/go/trajir/nodes"
+	"github.com/Coder-s-OG-s/Trajectory-IR/go/trajir/redact"
 )
 
 // Package format constants match the Python reference.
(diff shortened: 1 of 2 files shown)


jakharmonika364 (approved):
LGTM, quick bumping again

PR #382, opened 2026-09-17
fix: redact nested thoughts inside project context during tir export

### LINKED ISSUE #316: redact_payload doesn't catch THOUGHT nested inside PROJECT_CONTEXT
### Component 
`pkg/trajectory_ir/runtime/redact.py` `redact_payload(kind, payload)` only fully collapses a payload when the node's own `kind` is `THOUGHT`. A `PROJECT_CONTEXT` node embeds earlier nodes, including THOUGHT ones, inside its `items` list, and those get walked by the generic secret-key/value scrubber instead of being collapsed. `redact_projection_context()` already exists and handles this correctly per item, but `export_tir` never calls it.

```
# redact.py:48-60
def redact_payload(kind: str, payload: Any) -> dict[str, Any]:
    if kind == "THOUGHT":
        return {"redacted": True}
    if isinstance(payload, dict):
        return {k: redact_value(str(k), v) for k, v in payload.items()}
    return {"redacted": True}
```

### Steps to reproduce
1. Log a THOUGHT node with prose reasoning text (not matching `SECRET_VALUE_RE`).
2. Run `project_context()` so that THOUGHT node gets pulled into a `PROJECT_CONTEXT` node's `items` payload.
3. `export_tir(node_log, trajectory_id, "out.tir", redacted=True)`.
4. Unzip, find the `PROJECT_CONTEXT` node, inspect `payload.items[].payload`.

### Expected
- the embedded THOUGHT item's payload collapses to `{"redacted": true}`, same as if exported on its own.

### Actual
- raw THOUGHT text is still present inside the `PROJECT_CONTEXT` node in a "redacted" export.

### Suggested fix
- have `_redact_node` call `redact_projection_context` when `kind == "PROJECT_CONTEXT"` instead of the generic path.

Ayush4958:
## Summary

Fixes a data leak where nested `THOUGHT` payloads inside `PROJECT_CONTEXT` items escaped redaction during `.tir` export. Both the Python and Go SDKs have been patched to properly route projection contexts through their dedicated redaction handlers.

## Related issues

Closes #316

## How I tested

```bash
# Python
pytest test/unit/test_tir_package.py -q
pytest conformance/ -q
# Go
cd go && go test ./trajir/tir ./trajir/redact -count=1
```

## Checklist

## Safety areas

- [ ] No
- [x] Yes (Addresses secret handling and thought stripping heuristics during export)

## AI assistance (if any)

Test verification & Git Diff


Ayush4958:
@Siddhartha-singh01 @Ayush-Patel-56 
PTAL

thanks ...

Ayush-Patel-56 (approved):
Cheers! Looks Good

jakharmonika364 (commented):
fix checks out on both Go and Python side. But what I think is that the `PROJECT_CONTEXT` special-case is in each call site instead of `redact_payload` itself, so a future caller could skip it and reintroduce this. 

Still this is what I think for a follow-up, look into it. If feels good, appreciate a follow-up PR on it

jakharmonika364 (approved):
for now, this looks good to me. Thanks!

